PR #2, opened 2026-02-05
Major refactor and WIP DB support

jaredglaser:


jaredglaser:
@claude try again, the oauth token is set

PR #27, opened 2026-02-27
feat: Container history page with timeline navigation

jaredglaser:
## Summary
- Adds a per-container detail page (`/docker/$containerId`) for exploring historical metrics with interactive ECharts
- New `getDockerStatsForContainer` repository method with TimescaleDB `time_bucket` aggregation and configurable `targetPoints` for efficient data retrieval across any time range
- Draggable timeline slider with range presets (1h–30d), debounced chart updates, and URL-persisted state (metrics, from, to)
- History link icon on each container row for quick navigation
- Unit tests for new repository methods (`getDockerStatsForContainer`, `getContainerInfo`) and previously untested `getSourceIcons`

## Test plan
- [x] Verify `bun test` passes with new repository tests (15 new test cases)
- [x] Verify `bun run typecheck` passes
- [x] Navigate to Docker dashboard, hover a container row, click the history icon
- [x] Confirm charts render for default metrics (CPU, Memory) over the last hour
- [x] Toggle metric checkboxes on/off and verify charts update
- [x] Drag the timeline slider to select a sub-range and confirm charts update after debounce
- [x] Click range presets (1h, 6h, 12h, 1d, etc.) and verify timeline + charts refresh
- [x] Verify URL search params update when changing metrics or time range
- [x] Reload the page and confirm state restores from URL params
- [x] Test with `host` search param for multi-host setups

## Summary by CodeRabbit

* **New Features**
  * Container history page with interactive time-series charts, selectable metrics, consistent axis scaling, loading states, icon fallbacks, sticky timeline, and per-container route.
  * Preset time ranges (1h–30d), zoomable timeline, debounced updates, and URL sync for ranges/metrics.
  * Direct “History” link from container rows.

* **Tests**
  * Added unit tests covering historical metrics retrieval and repository behavior.

* **Chores**
  * Docs and migration updated: compression set to 7 days and retention wording adjusted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -310,7 +310,7 @@ migrations/                          # SQL migrations (settings + TimescaleDB wi
 - [x] **Docker Compose deployment** — multi-container setup with TimescaleDB, web server, and background worker
 - [x] **Database-backed streaming** — frontend reads from TimescaleDB via shared server-side polling instead of direct API/SSH connections; a `StatsPollService` runs 1 query/sec per source and broadcasts to all SSE clients
 - [ ] **Return to TanStack Start streaming server functions** — currently using SSE as a workaround because streaming server functions don't close quickly enough when rapidly switching between tabs; once TanStack Start's abort signal propagation is more reliable, migrate back to the native streaming pattern
-- [ ] **Historical data UI** — charts and graphs for historical metrics with time-range selection
+- [x] **Historical data UI** — per-container history page with time-bucketed charts, metric selection, and timeline navigation with range presets (1h–30d)
 - [x] **Proxmox API integration** — cluster overview with per-node VM, container, and storage monitoring via REST API polling
 - [ ] **Authentication** — user login and access control using OIDC with first class Pocket ID support
 - [ ] **Pre-built Docker image** — publish to a container registry for one-step deployment without building from source


jaredglaser (commented):
## Code Review: Container History Page

Nice feature overall — the architecture cleanly separates timeline vs. chart queries, the debounce pattern for slider drags is well-thought-out, and the URL param persistence makes the page bookmarkable. Here are my findings:

---

### Issues

#### 1. No error state UI (Medium)
`ContainerHistoryPage.tsx:205` — When `chartQuery` or `timelineQuery` fails, the user sees "No data available for this time range" which is indistinguishable from a genuinely empty range. Consider checking `chartQuery.isError` and showing a distinct error message.

```tsx
// Current
{chartData.length === 0 && !chartQuery.isFetching ? (
  <div>No data available for this time range.</div>
) : ( ... )}

// Suggested
{chartQuery.isError ? (
  <div>Failed to load chart data. Try refreshing.</div>
) : chartData.length === 0 && !chartQuery.isFetching ? (
  <div>No data available for this time range.</div>
) : ( ... )}
```

#### 2. `includesNow` recomputes stale `Date.now()` on every render (Low)
`ContainerHistoryPage.tsx:104` — `Date.now()` is called on every render, but `debouncedRange.to` only changes on debounce ticks. If the user is idle for minutes, `includesNow` could flip to `false` without the range actually changing, silently stopping auto-refresh. Consider wrapping this in a `useMemo` or using an interval to periodically re-evaluate.

#### 3. `handlePreset` suppression pattern is fragile (Low)
`HistoricalTimeline.tsx:178-186` — The `suppressZoomRef` is set to `true`, an action dispatched, then immediately set back to `false`. This relies on ECharts firing the `datazoom` event synchronously within `dispatchAction`. While this is currently true, it's an implicit coupling. A brief comment explaining why this works would help future maintainers.

#### 4. `handleMetricsChange` is a redundant wrapper (Nit)
`ContainerHistoryPage.tsx:155-157` — `handleMetricsChange` just forwards to `setSelectedMetrics`. You could pass `setSelectedMetrics` directly to `MetricCheckboxes.onChange` since React state setters are already stable references.

#### 5. `getGridClass` has identical branches (Nit)
`HistoricalChartsGrid.tsx:66-72` — `case 2` and `default` return the same value. Could simplify:
```tsx
function getGridClass(count: number): string {
  return count === 1 ? 'grid-cols-1' : 'grid-cols-1 xl:grid-cols-2';
}
```

---

### Positive Observations

- **Dual-query architecture**: Separating timeline (2000 points, coarse) from chart detail (600 points, fine-grained) with debounced transitions is a smart UX pattern. The timeline stays responsive while the charts load precise data.
- **URL state persistence**: Zod-validated search params with `z.coerce.number().pipe(z.number().finite())` is good defensive validation — prevents `Infinity`/`NaN` from reaching the DB queries.
- **`fromMs`/`toMs` swap in server function**: `docker.functions.tsx:79-80` — `Math.min`/`Math.max` ensures the range is always valid regardless of param order. Nice defensive coding.
- **Dynamic imports**: All server-only modules (`database-client`, `stats-repository`, etc.) are dynamically imported inside handlers, consistent with the project convention.
- **Parameterized SQL**: The `params.push(host)` pattern for optional host filters is clean and avoids any injection risk.
- **Test coverage**: 15 new test cases covering `getDockerStatsForContainer`, `getContainerInfo`, and `getSourceIcons` with edge cases (empty results, optional host filter, bucket size computation).
- **`@tanstack/react-query` as direct dependency**: Was previously imported (`ContainerRow.tsx` used `useQueryClient`) but only available transitively. Promoting it to a direct dependency is the right call.

---

### Questions / Minor Considerations

- **`Math.max(...values, 0)` with large datasets** (`HistoricalMetricChart.tsx:69`): With `targetPoints` capped at 5000, spreading into `Math.max` won't hit the call stack limit, but it's worth being aware of if that cap ever increases.
- **30-day range vs. 7-day retention**: The DB retention policy is 7 days per CLAUDE.md, but the timeline offers a 30d preset. Users selecting 30d will see mostly empty charts unless retention is extended. Is that intentional, or should the preset cap match the retention policy?

PR #119, opened 2026-04-19
fix(deploy): recover stuck deploys on startup + watchdog timeout

### LINKED ISSUE #99: Deploys stuck at `in_progress` after process crash are unrecoverable
## Problem

If the web server crashes between setting a deploy's status to `in_progress` (pipeline.ts line 118) and recording `succeeded`/`failed`, the deploy row is permanently stuck. The partial unique index (`idx_deploy_one_active_per_stack_host`) then rejects all future deploys for that stack/host combination.

There is no timeout-based cleanup, watchdog, or manual recovery path.

## Impact

A single process crash during deploy permanently blocks all future deploys for the affected stack/host until someone manually updates the DB.

## Suggested Fix

Options (not mutually exclusive):
1. **Timeout watchdog**: Background job that marks `in_progress` deploys older than X minutes as `failed`
2. **Startup recovery**: On server start, scan for `in_progress` deploys and mark them `failed` with a "recovered after crash" message
3. **Manual override**: Add a server function/UI action to force-cancel stuck deploys

## Files
- `src/lib/deploy/pipeline.ts:116-131`
- `src/lib/database/repositories/deploy-repository.ts` (partial unique index enforcement)

jaredglaser:
## Summary
- Add `DeployRepository.recoverStuckDeploys()` — called on server startup to mark any `pending`/`in_progress` rows as `failed`.
- Add `DeployRepository.timeoutStuckDeploys()` and `DeployWatchdog` — periodic job that fails deploys stuck past a threshold (default 30min, configurable).
- Prevents a single server crash or hung deploy from permanently blocking the stack+host via the `idx_deploy_one_active_per_stack_host` partial unique index.

Closes #99.

## Test plan
- [ ] `bun run typecheck` clean
- [ ] `bun test` green
- [ ] Startup recovery: simulate server with `in_progress` row → starts → row flipped to `failed` with recovery message
- [ ] Watchdog: simulate `in_progress` row older than threshold → watchdog tick flips it to `failed` with timeout message
- [ ] Log messages are draft text — reviewer should edit wording if needed

## Summary by CodeRabbit

* **New Features**
  * Added automatic detection and recovery of deployments stuck in progress
  * Introduced periodic watchdog monitoring to identify and handle stuck deployment states
  * Added automatic recovery of interrupted deployments during server startup
  * Configurable watchdog interval and timeout thresholds via optional environment variables

* **Tests**
  * Comprehensive test coverage added for watchdog and recovery mechanisms

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -43,8 +43,7 @@ WORKER_COLLECTION_INTERVAL_MS="1000"
 GIT_REPOS_DIR="./data/repos"
 GIT_SERVER_TOKEN="dev-git-token"
 
-# Deploy Watchdog — fails any deploy stuck in "in_progress" longer than the threshold.
-# Server startup separately flips pending/in_progress rows to "failed" to recover from crashes.
+# Deploy watchdog — fails deploys stuck in "in_progress" past the threshold.
 # DEPLOY_WATCHDOG_INTERVAL_MS="120000"
 # DEPLOY_WATCHDOG_TIMEOUT_MINUTES="10"
 
--- a/.env.example
+++ b/.env.example
@@ -43,6 +43,11 @@ WORKER_COLLECTION_INTERVAL_MS="1000"
 GIT_REPOS_DIR="./data/repos"
 GIT_SERVER_TOKEN="dev-git-token"
 
+# Deploy Watchdog — fails any deploy stuck in "in_progress" longer than the threshold.
+# Server startup separately flips pending/in_progress rows to "failed" to recover from crashes.
+# DEPLOY_WATCHDOG_INTERVAL_MS="120000"
+# DEPLOY_WATCHDOG_TIMEOUT_MINUTES="30"
+
 # Dev Agent (auto-seeded when DEV_AGENT_TOKEN is set)
 # DEV_HOST_NAME defaults to "localhost"; uncomment to override.
 # DEV_HOST_NAME="localhost"
--- a/.env.example
+++ b/.env.example
@@ -46,7 +46,7 @@ GIT_SERVER_TOKEN="dev-git-token"
 # Deploy Watchdog — fails any deploy stuck in "in_progress" longer than the threshold.
 # Server startup separately flips pending/in_progress rows to "failed" to recover from crashes.
 # DEPLOY_WATCHDOG_INTERVAL_MS="120000"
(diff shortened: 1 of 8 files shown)


jaredglaser:
## Reviewer notes from implementation agent

**Scope of startup recovery (worth considering now):**
`recoverStuckDeploys` currently flips **all** pending/in_progress rows on startup, without filtering by owning process. Fine for single-instance deployments (current state), but if we ever run multiple web processes against the same DB, one instance starting would fail in-flight deploys on another. Options if we want to pre-empt this:
- Scope by an owning instance id (would need a new column).
- Leave as-is and revisit when multi-process deploy is on the roadmap.

**Log message wording (drafts — please edit):**
- Startup recovery: _"Deploy interrupted — server restarted while this deploy was in progress. The actual outcome on the host is unknown. Please verify stack status and re-trigger if needed."_
- Watchdog timeout: _"Deploy marked failed by watchdog — no completion signal received after {minutes} minutes. The deploy may still be running on the host; check the agent before re-triggering."_

**Watchdog defaults** (env-overridable via \`DEPLOY_WATCHDOG_INTERVAL_MS\` / \`DEPLOY_WATCHDOG_TIMEOUT_MINUTES\`):
- Check interval: 2 min
- In-progress threshold: 30 min

**Test suite note:**
Full \`bun test\` shows 9 failures in \`DualSeriesChart\` / chart options / \`stacks-context\` — confirmed pre-existing on stock \`main\` (main: 2075 pass / 11 fail / 1 error). Unrelated to this PR; they pass in isolation.

jaredglaser:
## Follow-up: watchdog threshold should be tighter

30 min is too loose. The agent's own subprocess timeout (\`COMPOSE_TIMEOUT_MS\` in \`agent/src/routes/stacks.ts:7\`) is **5 min** — it kills the compose subprocess and returns a \`failed\` result at that point, which the pipeline records normally. So a row still \`in_progress\` after ~5 min means the failure path itself got lost (web process crashed, connection dropped mid-response, etc.).

Proposing: change the default \`DEPLOY_WATCHDOG_TIMEOUT_MINUTES\` from **30 → 10** (2× the agent's subprocess cap, leaves margin for startup/pull but catches genuinely-stuck rows quickly). Happy to push the change on this branch — let me know, or edit it directly in review.

jaredglaser:
@coderabbitai review

PR #120, opened 2026-04-19
fix(git): eliminate createStackInRepo race with callback-style commitFiles

jaredglaser:
## Summary
- Refactored `commitFiles` to accept a callback that runs INSIDE `withRepoLock` and receives a `ReadonlyMap<string, string>` snapshot of the current HEAD tree. The callback returns a `CommitPlan` derived from that fresh state, so stale-read races like #98 become impossible by construction.
- Updated all callers: `createStackInRepo`, `deleteStackFromRepo`, `saveAndCommitFile`, `updateManifest`, and `ensureRepoInitialized`. Host validation and active-deploy checks stay outside the git lock to avoid holding the mutex across PostgreSQL round-trips.
- Added a concurrent-create race test plus callback-semantics coverage in `repo.test.ts`.

Closes #98.

## Test plan
- [x] `bun test` — 2107 pass, 0 new failures (9 pre-existing failures in `DualSeriesChart`, `getChartOption`, `stacks-context` are unrelated to this change)
- [x] New concurrent-commit race test covers the issue-#98 scenario (two parallel manifest-merging commits both land without clobber)
- [x] New coverage for callback snapshot semantics: receives prior HEAD contents, empty map on first commit, errors propagate without writing a commit

Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved atomicity of file operations by enabling functions to inspect current state before committing changes, reducing race conditions in concurrent scenarios.

* **Tests**
  * Updated test infrastructure to verify new concurrency handling for stack and manifest operations.

### DIFF
--- a/src/lib/git/__tests__/editor-operations.test.ts
+++ b/src/lib/git/__tests__/editor-operations.test.ts
@@ -13,14 +13,14 @@ describe('saveAndCommitFile', () => {
     testDir = mkdtempSync(join(getTestTmpDir(), 'git-editor-'));
     repoPath = join(testDir, 'test.git');
     await initBareRepo(repoPath);
-    await commitFiles(repoPath, {
+    await commitFiles(repoPath, () => ({
       files: [
         { path: 'manifest.yaml', content: 'stacks:\n  plex:\n    host: homeserver\n    autoDeploy: true\n' },
         { path: 'plex/docker-compose.yml', content: 'services:\n  plex:\n    image: plex:v1\n' },
       ],
       message: 'initial',
       author: { name: 'system', email: 'system@localhost' },
-    });
+    }));
   });
 
   afterEach(() => {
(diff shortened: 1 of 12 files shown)


jaredglaser:
## Reviewer notes from implementation agent

**API is a clean break** — old direct-plan \`commitFiles(path, { files, message, author })\` no longer exists. All 4 callers (\`createStackInRepo\`, \`deleteStackFromRepo\`, \`saveAndCommitFile\`, \`updateManifest\`) now pass a callback. If an external caller exists outside this repo (not expected), it will get a typecheck failure rather than a silent race.

**Race test location caveat** (worth knowing):
The agent originally placed the race test at the \`createStackInRepo\` layer in \`stack-service.test.ts\`, but it failed in full-suite runs because pre-existing component tests (\`ComposeEditor*.test.tsx\`, \`DeployHistoryList.test.tsx\`) do \`mock.module('@/lib/stacks/stack-service', ...)\` which pollutes globally across Bun's concurrent test runner — the CLAUDE.md anti-pattern, but pre-existing code not touched here. The race test was moved down to \`repo.test.ts\` where it exercises the same callback-inside-lock invariant directly via \`commitFiles\`. **This is the right fix for the race, but the pre-existing \`mock.module()\` pollution is still a latent footgun — separate cleanup issue worth filing.**

**Test suite note:**
Full \`bun test\` shows 9 failures in \`DualSeriesChart\` / chart options / \`stacks-context\` — confirmed pre-existing on stock \`main\`. Unrelated to this PR.

PR #122, opened 2026-04-19
fix(stats-poll): refresh DB pool reference each tick (#102)

jaredglaser:
## Problem (#102)

`StatsPollService` in `src/lib/database/subscription-service.ts` cached a single `StatsRepository` instance on `this.repo`. That repo held a reference to the `pg.Pool` from the *initial* `databaseConnectionManager.getClient()` call.

When `DatabaseClient`'s pool `'error'` handler fires (e.g. the DB container restarts), it flips `connected=false`. The next `getClient()` call rebuilds a fresh `DatabaseClient` + `Pool`. But `statsPollService.repo` still held the old, dead pool — every poll tick queried a pool whose connections were never coming back, and the whole stats SSE stream silently stopped delivering rows until the process restarted.

## Fix

Rebuild the repo per poll tick instead of caching it:

```ts
const config = loadDatabaseConfig();
const dbClient = await databaseConnectionManager.getClient(config);
const repo = new StatsRepository(dbClient.getPool());
```

`databaseConnectionManager.getClient()` is already cached on the healthy path — it returns the existing `DatabaseClient` from its `Map` when `isConnected()` is true. So the per-tick cost is one `Map.get` + one `new StatsRepository(pool)` (a constructor that only stores a reference). Negligible overhead, and the poll service now transparently recovers after any pool reconnect.

Also removed the now-dead `private repo` field and `getRepo()` method, and dropped `this.repo = null` from `stop()`.

## Tests

Added `src/lib/database/__tests__/subscription-service.test.ts` covering:

1. **Per-tick rebuild uses current client** — two successive `getClient()` calls return two different clients holding two different pools; verifies the repo built on tick 2 holds the second pool (spying on `StatsRepository.prototype.getDockerStatsSince` and reading `this.pool`).
2. **Healthy path** — same client returned twice; two ticks run without error and don't log.
3. **`getClient` throws** — existing `try/catch` still handles it; `consecutiveFailures` increments; `onError` fires exactly once at the threshold and doesn't re-fire within the same episode.

Tests use `spyOn` on the `databaseConnectionManager` singleton and on `setInterval`/`clearInterval` — no `mock.module()`, so no global pollution.

## Verification

- `bun run typecheck` passes
- `bun test src/lib/database/__tests__/subscription-service.test.ts` → 3/3 pass
- Full `bun test`: +3 passing tests vs. `origin/main`, same 8 pre-existing unrelated failures (chart tests that fail under parallel-run pollution from other `mock.module()` callers — unchanged by this PR)

Closes #102

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test suite for stats polling service covering connection recovery, error handling, and polling behavior.

* **Refactor**
  * Improved database connection acquisition in stats polling service for enhanced resilience during polling cycles.

### DIFF
--- a/src/lib/database/__tests__/subscription-service.test.ts
+++ b/src/lib/database/__tests__/subscription-service.test.ts
@@ -1,0 +1,167 @@
+import { describe, it, expect, beforeEach, afterEach, spyOn } from 'bun:test';
+import { databaseConnectionManager } from '@/lib/clients/database-client';
+import { StatsRepository } from '@/lib/database/repositories/stats-repository';
+import { statsPollService } from '../subscription-service';
+
+interface IntervalHandle {
+  cb: () => Promise<void> | void;
+  ms: number;
+  cleared: boolean;
+}
+
+/**
+ * Replace setInterval/clearInterval so each registered interval becomes a
+ * callback we can invoke deterministically from tests (no real timers run).
+ */
+function createIntervalHarness() {
+  const intervals: IntervalHandle[] = [];
+  const setSpy = spyOn(globalThis, 'setInterval').mockImplementation(((cb: () => void, ms: number) => {
+    const handle: IntervalHandle = { cb, ms, cleared: false };
+    intervals.push(handle);
+    return handle as unknown as ReturnType<typeof setInterval>;
+  }) as typeof setInterval);
+  const clearSpy = spyOn(globalThis, 'clearInterval').mockImplementation(((h: unknown) => {
+    const handle = h as IntervalHandle;
+    if (handle) handle.cleared = true;
+  }) as typeof clearInterval);
+  return {
+    intervals,
+    restore() {
+      setSpy.mockRestore();
+      clearSpy.mockRestore();
+    },
+  };
+}
+
+/** Minimal pg.Pool stand-in that records which pool instance received queries. */
+function createMockPool(label: string) {
+  const queries: { sql: string; params: unknown[] }[] = [];
+  return {
+    label,
+    queries,
+    pool: {
+      query: async (sql: string, params: unknown[] = []) => {
+        queries.push({ sql, params });
+        return { rows: [] };
+      },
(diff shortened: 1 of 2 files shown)


jaredglaser:
## PR Review

**Summary:** Correctly fixes a reliability bug where `StatsPollService` held a stale `StatsRepository` after a DB reconnect, causing the stats stream to silently fail until process restart. The per-tick rebuild is the right approach since `databaseConnectionManager.getClient()` already absorbs the cache hit on the healthy path.

**Issues Found:**
- [`src/lib/database/__tests__/subscription-service.test.ts`:93] The reconnect test reads `StatsRepository`'s private `pool` field via a double type cast (`this as unknown as { pool: import('pg').Pool }`). TypeScript cannot catch a rename of that field because the `unknown` cast bypasses the type system. If `StatsRepository.pool` is ever renamed, this assertion silently stops testing anything meaningful. Consider asserting on `getClientSpy` call count or which mock pool received the query (`firstPool.queries.length` / `secondPool.queries.length`) rather than reading the private field directly.

**Suggestions:**
- [`src/lib/database/subscription-service.ts`:71] `loadDatabaseConfig()` runs a Zod schema parse + `process.env` reads on every 1-second tick. The config is static at runtime; hoisting it to be computed once at construction time (or lazily cached) avoids repeated parse overhead. The `databaseConnectionManager.getClient(config)` call is already cached on the healthy path, so the config object is the only unnecessary work each tick.

**Looks Good:**
- Core fix is sound: removing the cached `this.repo` / `getRepo()` pattern and rebuilding `StatsRepository` per tick correctly picks up a fresh pool after any reconnect.
- `stop()` is correctly simplified — `this.repo = null` removal matches the field deletion.
- All three tests cover the key contracts (per-tick rebuild, healthy no-op, failure threshold) with narrow prototype spies rather than `mock.module()` — aligned with project test isolation guidelines.
- No `console.log` violations; error logging preserved correctly.

PR #123, opened 2026-04-19
refactor(utils): extract shared backoff utility and migrate 4 call sites

jaredglaser:
## Summary

Extracts duplicated exponential-backoff math into `src/lib/utils/backoff.ts` and migrates call sites to use the shared helper. Zero behavior change for the original 3 call sites — delay sequences are byte-identical before and after. A 4th call site (`settings-broadcast-service`) was added in a follow-up commit and involves a minor behavior change (see below).

## New module

- `src/lib/utils/backoff.ts`
  - `backoffDelayMs(attempt, opts?)` — pure calculator. `delay = min(baseMs * 2^min(attempt, maxExponent), capMs)`. Defaults: `baseMs=500`, `capMs=30_000`, `maxExponent=Infinity`. Negative attempts clamped to 0.
  - `retry(fn, opts)` — runs `fn` with exponential backoff between attempts; respects an abort signal (sleep rejects with AbortError on abort); non-retryable errors short-circuit.
- `src/lib/utils/__tests__/backoff.test.ts` — full coverage of the table in the task spec plus retry semantics (first-attempt success, mid-sequence success, all-fail, non-retryable short-circuit, `maxAttempts: 1`, already-aborted signal, invalid `maxAttempts: 0`).

## Migrated call sites (3)

1. **`src/lib/docker/docker-inventory-broadcast-service.ts`** — local `backoffDelay()` removed; both call sites now use `backoffDelayMs(this.reconnectFailures, { baseMs: BACKOFF_BASE_MS, capMs: BACKOFF_CAP_MS })`. `BACKOFF_CAP_MS = 32_000` preserved, `maxExponent` omitted → `Infinity`, same cap-bounded sequence.
2. **`src/lib/stacks/stack-status-broadcast-service.ts`** — same pattern as (1).
3. **`src/worker/collectors/base-collector.ts`** — in-line `Math.min(Math.pow(...))` replaced with `backoffDelayMs(this.consecutiveErrors, { baseMs: BASE_BACKOFF_MS, capMs: MAX_BACKOFF_MS, maxExponent: MAX_BACKOFF_EXPONENT })`. Constants unchanged; they now flow through as options.

## Now also migrates

- **`src/lib/settings/settings-broadcast-service.ts`** — previously used a fixed `5_000`ms reconnect delay. Now uses `backoffDelayMs(this.reconnectFailures, { baseMs: 500, capMs: 30_000 })` with the counter reset on successful LISTEN, matching the pattern in docker-inventory and stack-status. **Minor behavior change**: the first reconnect attempt is now 500ms (instead of 5000ms — faster recovery from transient blips), and sustained outages back off up to 30s (instead of hammering every 5s). The reconnect trigger conditions (listener error handler, initial-subscribe connect failure) are unchanged. DB access is now injected via constructor deps so the reconnect cadence can be unit-tested without touching PostgreSQL; `src/lib/settings/__tests__/settings-broadcast-service.test.ts` adds 8 new tests covering init/change/broadcast and the backoff sequence (first failure = 500ms, exponential growth, cap at 30_000, reset after successful reconnect).

## Not migrated

- **`src/lib/hosts/host-utils.ts` (`retryHealthCheck`)** — uses a fixed delay array (`HEALTH_CHECK_DELAYS_MS`). Different pattern, intentionally skipped.
- **`src/hooks/useEventSource.ts`** — browser-side; no `abortableSleep` (which depends on `AbortSignal` + `setTimeout` teardown patterns that the hook already handles differently). Intentionally skipped.

## Why

Today the same `Math.min(BASE * 2^n, CAP)` formula is copy-pasted into 3+ services. Consolidating gives one place to reason about the delay curve and lets the forthcoming `AgentClient` retry (PR #100) share the same helper. Aligning `settings-broadcast-service` on the same utility also means all three reconnecting broadcast services recover on the same cadence.

## Test plan

- [x] `bun run typecheck` — clean.
- [x] `bun test src/lib/utils/__tests__/backoff.test.ts` — new 12 tests pass.
- [x] `bun test src/lib/settings/__tests__/settings-broadcast-service.test.ts` — new 8 tests pass.
- [x] `bun test` targeted at migrated modules — pre-existing tests continue to pass unchanged (delay assertions still see `500` → `1000` → `...`).
- [x] `bun test` full suite — same pass/fail count as main baseline (9 pre-existing flaky failures unrelated to this change, in DualSeriesChart/stacks-context tests that pass in isolation).


jaredglaser:
## PR Review

**Summary:** Cleanly extracts duplicated exponential backoff logic into a shared `src/lib/utils/backoff.ts` utility and migrates three broadcast services and `BaseCollector` to use it. `SettingsBroadcastService` gets a meaningful behavioral improvement (fixed 5s delay replaced with proper exponential backoff) along with DI seams for testability.

**Issues Found:**
- [`src/lib/utils/backoff.ts`, `retry()` loop body] `const signal = opts.signal ?? new AbortController().signal` allocates a fresh `AbortController` on every retry iteration when no signal is provided. Move it above the loop so a single throwaway controller is shared across retries.
- [`src/lib/deploy/startup-recovery.ts`] Not touched by this PR, but contains its own private `runWithRetry` helper with an overlapping interface to the new `retry()` — leaving a 5th independent backoff implementation behind. Not a blocker, but the consolidation motivation applies here too.
- [`src/lib/docker/docker-inventory-broadcast-service.ts`, `src/lib/stacks/stack-status-broadcast-service.ts`] Both use `capMs: 32_000` while `settings-broadcast-service.ts`, `base-collector.ts`, and the `backoff.ts` default all use `30_000`. Worth aligning or adding a comment explaining the intentional difference.

**Suggestions:**
- [`src/lib/utils/backoff.ts`, `retry()`] The final `throw lastErr` at the bottom of the function is unreachable in practice (the last iteration always throws from within the loop). Consider removing it or restructuring so the invariant is clear and TypeScript strict mode does not require `lastErr` initialized before the loop.
- Consider migrating `startup-recovery.ts`'s `runWithRetry` to the new shared `retry()` in a follow-up, or add a TODO comment linking them.

**Looks Good:**
- `backoffDelayMs()` is a clean pure function; `maxExponent` as an opt-in parameter handles `BaseCollector`'s exponent cap elegantly.
- DI approach in `SettingsBroadcastService` avoids `mock.module()` on barrel modules — correct per CLAUDE.md.
- `setTimeout` spy scoped to the nested `reconnect backoff` describe block with proper `beforeEach`/`afterEach` teardown — follows CLAUDE.md rule 7.
- All logging uses `console.error`/`console.info`; no drive-by `console.log` statements.
- Dynamic `await import()` pattern preserved throughout `settings-broadcast-service.ts`.
- Behavioral parity confirmed for all three existing call sites: same constants, same formula.

PR #124, opened 2026-04-19
feat(agent): retry transient failures in AgentClient (#100)

jaredglaser:
## Problem

`AgentClient.request()` throws `AgentClientError` on the very first failure (any network error or non-2xx response). The deploy pipeline catches that and marks the deploy `failed` permanently. A brief network hiccup, agent restart, or transient 503 mid-deploy is therefore unrecoverable.

## Fix

Wrap `AgentClient.request()` with the shared `retry()` helper (from PR #123). Transient failures retry up to 3 total attempts with exponential backoff (1s, 2s). Non-transient failures still fail immediately.

### Retry policy

- **Retryable:** fetch throws that are NOT our own per-attempt timeout (DNS, ECONNREFUSED, ECONNRESET, socket hangup) + HTTP 502, 503, 504.
- **Non-retryable (fail immediately):** all 4xx; all 5xx other than 502/503/504; our own per-attempt timeout firing; invalid JSON bodies.
- **Attempts:** 3 total (1 initial + 2 retries).
- **Backoff:** `retry()` with `baseMs: 1000`, `maxExponent: 2` → sleeps of 1000ms, 2000ms.
- **Per-attempt timeout:** unchanged — each attempt still uses `AbortSignal.timeout(this.timeoutMs)`. There is no outer total budget. Because timeouts are treated as non-retryable, the per-attempt timeout does not compound.

### Key safety decision

Per-attempt timeouts are never retried. A `docker compose up` that locally times out may still be running on the agent; firing a second attempt would double-apply. This is enforced via a new `AgentClientError.wasTimeout` field (defaulted `false` — backward-compatible) that the retry classifier reads.

## Scope

- `streamZfsStats` is NOT retried (it has its own fetch path; the ZFS collector handles reconnect via `BaseCollector`'s backoff).
- No pipeline changes.
- No env-var configuration.
- No jitter.

## Tests

New `retry behaviour` block in `agent-client.test.ts` covering:
1. Success on first attempt → 1 fetch, no sleep.
2. Network error × 2, then success → 3 fetches, delays `[1000, 2000]`.
3. 503 then 200 → 2 fetches, delay `[1000]`.
4. 400 → throws immediately, 1 fetch.
5. 500 (non-retryable 5xx) → throws immediately, 1 fetch.
6. 504 × 3 → throws with `statusCode: 504`, 3 fetches.
7. Network error × 3 → throws with `statusCode: undefined`, 3 fetches.
8. Per-attempt timeout (`TimeoutError`) → `wasTimeout: true`, 1 fetch.
9. `AbortError` variant → `wasTimeout: true`, 1 fetch.
10. Invalid JSON (2xx with bad body) → throws immediately, 1 fetch.

`setTimeout` is spied at the block scope so sleeps fire synchronously.

Two pre-existing tests that rejected once now use `mockRejectedValue` (all attempts) since the new retry means a single rejection is no longer the terminal failure.

## Depends on

- #123 (shared backoff utility) — this PR's base is `reliability/backoff-utility`, not `main`. Should be rebased onto `main` and merged after #123.

## Verification

- `bun run typecheck` — clean.
- `bun test src/lib/clients/__tests__/agent-client.test.ts` — 26 pass.
- `bun test` — full suite passes (9 pre-existing unrelated failures in chart/stacks-context tests, confirmed present on base branch).

Closes #100

### DIFF
--- a/src/lib/clients/__tests__/agent-client.test.ts
+++ b/src/lib/clients/__tests__/agent-client.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, beforeEach, mock } from 'bun:test';
+import { describe, it, expect, beforeEach, afterEach, mock, spyOn } from 'bun:test';
 import { AgentClient, AgentClientError, type ZfsPool } from '../agent-client';
 
 describe('AgentClient', () => {
@@ -75,7 +75,8 @@ describe('AgentClient', () => {
     });
 
     it('throws AgentClientError on fetch failure (network error)', async () => {
-      fetchMock.mockRejectedValueOnce(new Error('Connection refused'));
+      // Network errors retry up to 3 times — reject all attempts so we see the final error.
+      fetchMock.mockRejectedValue(new Error('Connection refused'));
 
       await expect(client.deploy({
         stack: 'plex',
@@ -139,7 +140,8 @@ describe('AgentClient', () => {
     });
 
     it('throws AgentClientError when agent is unreachable', async () => {
-      fetchMock.mockRejectedValueOnce(new Error('ECONNREFUSED'));
+      // Network errors retry up to 3 times — reject all attempts.
+      fetchMock.mockRejectedValue(new Error('ECONNREFUSED'));
 
       await expect(client.health()).rejects.toThrow(AgentClientError);
     });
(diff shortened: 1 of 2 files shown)


jaredglaser:
## PR Review

**Summary:** Adds 3-attempt exponential backoff retry to `AgentClient.request()` for transient failures (network errors, 502/503/504), with per-attempt timeout detection preventing retry of in-flight `docker compose up` operations. Clean implementation with good test coverage.

**Issues Found:**
- [`src/lib/clients/agent-client.ts`:207] Hardcoded `/2` denominator in the retry log message: `Retry ${attempt}/2`. This is correct today (`maxAttempts: 3` → 2 retries), but if `maxAttempts` changes the log will silently produce misleading output like `Retry 3/2`. Consider computing it: capture `const maxRetries = 2` alongside `maxAttempts: 3` and reference it in `onRetry`.

**Looks Good:**
- `console.info` used for retry telemetry — correct per logging guidelines
- `setTimeout` spy correctly scoped to the `retry behaviour` describe block with `beforeEach`/`afterEach` — matches CLAUDE.md testing guidelines
- `AbortError` + `TimeoutError` both flagged as `wasTimeout: true` — safe since `attempt()` only uses `AbortSignal.timeout()` (no external signal reaches the `request()` path)
- Retry classifier is precise: network errors retry, 4xx/500/501/5xx-non-gateway fail immediately, 502/503/504 retry
- `mockRejectedValueOnce` → `mockRejectedValue` in the network error tests is correct — retry now makes multiple fetch calls
- Uses injectable `fetchFn` DI seam throughout — no `mock.module()` needed
- `@/lib/utils/backoff` import follows the `@/` path convention

PR #125, opened 2026-04-19
feat(teardown): async teardown via pipeline postSuccess hook (#101)

jaredglaser:
## Problem

`teardownAndAwait` in `src/lib/stacks/teardown-poller.ts` polled the deploy status for up to **120 seconds**, blocking the HTTP server-function response. The UI was frozen for that duration, and reverse proxies could time the request out before the teardown completed. (Issue #101.)

## Solution

The server function now returns a `deployId` immediately; the deploy pipeline owns the post-completion manifest delete via a new **`postSuccess` hook** on `DeployRequest`. The existing stack-status SSE stream and `deploy_change` NOTIFY drive the UI update when the agent finishes.

### Pipeline

- New optional `postSuccess?: 'removeFromManifest'` field on `DeployRequest`, persisted in `deploy_history.post_success`.
- After a successful dispatch (agent returns `success: true`), the pipeline invokes `stackRepoWriter.removeStackFromManifest(stack)`.
- If the manifest write fails, the deploy row is flipped to `failed` with an explanatory log message and the result returns `{ status: 'failed', ... }` so the UI can surface it.
- `no_change` also counts as success (teardown of a stack with nothing running).
- Missing `postSuccess` on the request short-circuits — `removeStackFromManifest` is never called.

### Stack service

- `deleteStackFromRepo` now returns a discriminated union:
  ```ts
  | { status: 'removed'; commitSha: string }           // unmanage (sync)
  | { status: 'teardown-pending'; deployId: number }    // teardown (async)
  ```
- Teardown branch calls `triggerStackDeploy({ action: 'teardown', postSuccess: 'removeFromManifest' })` and returns immediately. The pipeline handles the manifest delete asynchronously.
- Pure branch logic extracted to `src/lib/stacks/delete-stack-resolver.ts` so unit tests can exercise it in isolation (the `stack-service` module is heavily mocked by other tests).

### Startup recovery

- New orphaned-manifest-delete sweep added to `performStartupRecovery`.
- Calls `deployRepo.findSucceededPostSuccessDeploys('removeFromManifest')` and, for each returned stack that's still present in `manifest.yaml`, runs `stackRepoWriter.removeStackFromManifest(stack)` idempotently.
- Deduplicates multiple succeeded rows for the same stack; survives per-row failures; starts the watchdog even if the sweep throws.

### UI

- `DeleteStackDialog` → `handleDeleteConfirm` closes the dialog and navigates to `/stacks` synchronously; the mutation runs in the background.
- The existing Jotai `toastAtom` shows a "Teardown queued for X — the stack will be removed when the agent finishes." message. This persists across the navigation unmount (unlike the per-route `Snackbar`).

## Schema change

`migrations/020_add_deploy_post_success.sql`:

```sql
ALTER TABLE deploy_history ADD COLUMN post_success TEXT NULL;
```

Backward-compatible. Existing rows read as `postSuccess: null`.

## Migration for existing deployments

1. Apply migration `020_add_deploy_post_success.sql`.
2. No code changes required in downstream consumers — the old synchronous `teardownAndAwait` behaviour is replaced. UI callers receive the new discriminated union.
3. On first boot after upgrade, the orphaned-sweep runs once against any pre-existing rows (there won't be any with `postSuccess` set yet, so it's a no-op).

## Tests

- **Pipeline**: new `postSuccess hook` suite covering happy path, `no_change` triggers hook, manifest-delete throws, deploy failure skips hook, missing `postSuccess` skips hook.
- **DeployRepository**: `insertDeploy` persists `post_success`; `findSucceededPostSuccessDeploys` returns matching rows.
- **Startup recovery**: sweeps orphaned rows, skips already-gone stacks, continues after per-stack failure, deduplicates multi-row rows, skips when deps absent, watchdog starts even on sweep throw.
- **`resolveDeleteStack`**: both branches of the discriminated result.
- **`functions.test.ts`**: `mock.module('@/lib/stacks/stack-service', …)` now provides all exports including `resolveDeleteStack`, `deleteStackFromRepo`, `createStackInRepo`, `getManagedHostNames` — prevents downstream test pollution.

### Test results

`bun run typecheck` — clean.
`bun test` — 2150 pass / 9 fail. The 9 failures are **pre-existing** on `origin/main` (DualSeriesChart + stacks-context defaults) and unrelated to this PR.

## Non-goals / follow-up

- **Stack-status broadcaster gap** — verified that `deploy_change` NOTIFY is emitted by the pipeline after the manifest delete, and the existing `useStackStatus` hook bumps `deployVersion`, which invalidates the `STACKS_QUERY_KEY` and re-fetches `listStacks()` (which reads the manifest). The torn-down stack row disappears from the list naturally.
- The watchdog's timeout behaviour is preserved — remains the safety net for stuck deploys.
- No changes to agent protocol or SSE endpoints.

## Test plan

- [ ] Apply migration `020_add_deploy_post_success.sql`.
- [ ] Trigger "Tear down & remove" on a running stack.
  - [ ] Dialog closes immediately.
  - [ ] Navigation to `/stacks` is instant.
  - [ ] Toast "Teardown queued for X" appears.
  - [ ] After agent completes, stack row disappears from the list.
  - [ ] `deploy_history.post_success = 'removeFromManifest'` on the row; status reaches `succeeded`.
- [ ] Tear down a stack with no running containers. Row still gets removed (agent returns success fast).
- [ ] Simulate manifest-write failure (e.g., repo permissions). Verify deploy row flips to `failed` with "Teardown succeeded but manifest delete failed" logs.
- [ ] Kill the server between agent ACK and manifest delete. Restart. Verify the orphaned-sweep log appears and the manifest entry is cleaned up.
- [ ] Unmanage path (`teardown=false`) still returns `{ status: 'removed' }` synchronously.


jaredglaser:
## PR Review

**Summary:** Replaces the blocking 120-second teardown poll with an async pipeline postSuccess hook, returning a `deployId` immediately and delegating manifest cleanup to a post-dispatch callback with crash recovery via startup sweep. The architecture is sound and coverage is thorough.

**Issues Found:**
- [`src/lib/deploy/pipeline.ts`, no_change branch] The postSuccess hook invocation in the no_change path inside `execute()` is unreachable dead code. The no_change guard only fires for `action === 'deploy'` requests, but `postSuccess: 'removeFromManifest'` is only ever set when `action: 'teardown'` (via `resolveDeleteStack`). Teardown requests skip change detection entirely — so this branch can never be reached with `postSuccess` set. The no_change+postSuccess path has no test coverage because no test can construct a request satisfying both conditions. Consider removing the hook invocation from the no_change path or adding a comment explaining intentional future-proofing.
- [`src/data/stacks/__tests__/functions.test.ts`] `mockResolveDeleteStack` is declared and included in the `mock.module('@/lib/stacks/stack-service', ...)` call but no test exercises it. The `deleteStack` server function's `teardown-pending` return path (when the agent queues a teardown and returns `{ status: 'teardown-pending', deployId }`) is not covered at the `functions.tsx` integration level. Since all exports must be provided when mocking a service module, the mock entry is correct — but the missing test case leaves that union branch unverified.

**Suggestions:**
- [`src/lib/stacks/stack-service.ts`] `resolveDeleteStack` is re-exported from `stack-service.ts` (originating in `delete-stack-resolver.ts`). Any test importing it directly from `@/lib/stacks/delete-stack-resolver` would bypass the service module mock. Fine for now (no such tests exist), but the re-export boundary is slightly ambiguous — a brief JSDoc explaining it's exposed here for co-location with service-layer mocks would clarify intent.

**Looks Good:**
- All server-only modules in `server-init.ts` use `await import()` — no static leakage into the client bundle.
- `postSuccess` is persisted to the database *before* the hook executes — idempotent crash recovery via the startup sweep is correctly ordered.
- `toDeployRecord()` wraps `id` with `Number()` — BIGINT string coercion handled correctly.
- The startup sweep correctly deduplicates multiple rows per stack and skips stacks already absent from the manifest (idempotent).
- Jotai-backed `showToast` persists across route changes, so the error toast in `onError` remains visible even after `navigate({ to: '/stacks' })` fires optimistically.
- Migration is backward-compatible (`ALTER TABLE ... ADD COLUMN ... NULL`).

PR #127, opened 2026-04-19
refactor(db): add databaseMiddleware to eliminate DB setup boilerplate (#106)

### LINKED ISSUE #106: Add database middleware to eliminate repeated setup boilerplate
## Problem

Every stats-related server function handler repeats the same 6-line DB setup:

```typescript
const { databaseConnectionManager } = await import('@/lib/clients/database-client');
const { loadDatabaseConfig } = await import('@/lib/config/database-config');
const { StatsRepository } = await import('@/lib/database/repositories/stats-repository');
const config = loadDatabaseConfig();
const dbClient = await databaseConnectionManager.getClient(config);
const repo = new StatsRepository(dbClient.getPool());
```

This appears 10+ times across `docker/functions.tsx`, `zfs/functions.tsx`, `proxmox/functions.tsx`, `stacks/functions.tsx`, and `stack-service.ts`. Only `settings/functions.tsx` extracts a helper.

## Suggested Fix

Create a `databaseMiddleware` (similar to `openBaoMiddleware`) that injects the DB pool or a pre-built repository into the server function context:

```typescript
export const databaseMiddleware = createMiddleware().server(async ({ next }) => {
  const { databaseConnectionManager } = await import('@/lib/clients/database-client');
  const { loadDatabaseConfig } = await import('@/lib/config/database-config');
  const config = loadDatabaseConfig();
  const dbClient = await databaseConnectionManager.getClient(config);
  return next({ context: { pool: dbClient.getPool() } });
});
```

## Files
- New: `src/middleware/database-middleware.ts`
- `src/data/docker/functions.tsx`
- `src/data/zfs/functions.tsx`
- `src/data/proxmox/functions.tsx`
- `src/data/stacks/functions.tsx`
- `src/data/settings/functions.tsx`

jaredglaser:
## Summary
- Adds `src/middleware/database-middleware.ts` that injects a `pg.Pool` into the server-function context, mirroring the `openBaoMiddleware` pattern. Server-only modules (`database-client`, `database-config`) are dynamically imported **inside** the `.server()` callback, so they stay out of the client bundle.
- Validated config is cached in a module-scoped `cachedConfig` on first request (load-once semantics without top-level `await`). Bootstrap failures (`ZodError` on bad env, DB unreachable) are logged with a `[databaseMiddleware]` prefix and re-thrown. A `resetDatabaseMiddlewareState()` helper is exported for test isolation.
- Migrates 8 server functions across `docker`, `zfs`, `proxmox`, and `settings` `functions.tsx` to use the middleware. Each handler now reads `context.pool` and constructs the needed repository inline, eliminating ~9 lines of boilerplate per copy.
- Drops the inline `getSettingsRepository` helper in `settings/functions.tsx` — the middleware replaces it.
- Adds `src/middleware/__tests__/database-middleware.test.ts` covering happy-path injection, error-path rejection (with logging + `next` not called), cached-config reuse across requests, and context-shape pinning. Uses `spyOn(databaseConnectionManager, 'getClient')` rather than `mock.module()` to avoid global pollution (per CLAUDE.md rule #7).

### Migrated server functions
- `src/data/docker/functions.tsx` — 5 functions (`getHistoricalDockerStats`, `getDockerEntityIcons`, `getContainerHistory`, `getContainerInfo`, `updateContainerIcon`)
- `src/data/zfs/functions.tsx` — 1 (`getHistoricalZFSStats`)
- `src/data/proxmox/functions.tsx` — 1 (`getHistoricalProxmoxStats`)
- `src/data/settings/functions.tsx` — 1 (`updateSetting`)

### Behavior change (intentional)

DB-bootstrap failures (connection refused, auth failure, bad env) in the 6 handlers that previously had `try/catch` (`getHistoricalDockerStats`, `getHistoricalZFSStats`, `getHistoricalProxmoxStats`, `getDockerEntityIcons`, `getContainerHistory`, `getContainerInfo`) now reject the RPC instead of falling through to the old `catch` → `return []`/`{}`/`null` fallback. The catch block is now dead code for bootstrap errors (handler never runs if the middleware throws); it still catches repository-method errors that occur after `context.pool` is injected.

Surfacing DB outages beats the previous silent empty fallbacks. Preload queries will flip to `error` state in TanStack Query instead of rendering "empty, filling up". Deliberate; see `[databaseMiddleware] DB bootstrap failed` log prefix for ops triage.

### Non-goals
- `stacks/functions.tsx` has no DB-setup boilerplate (uses service-layer functions); its cached OpenBao client is left alone (tracked separately in #107).
- `hosts/functions.tsx` uses a different repository (`HostRepository`) via its own `loadDeps` helper — left as-is to keep this PR scoped to the stats/settings pattern called out in #106.

Closes #106

## Test plan
- [x] `bun test` — 2121 pass / 10 fail / 1 error. Failure set is identical to `main` baseline (pre-existing `DualSeriesChart` and `useSettings` flakes). Middleware file: 4/4 pass.
- [x] Middleware tests cover: server handler presence, happy-path context injection, context-shape pinning, `getClient` rejection (rejects + no `next` call + `console.error` with `[databaseMiddleware]` prefix), and cached-config reuse across two invocations.


jaredglaser:
## PR Review

**Summary:** Introduces `databaseMiddleware` to centralize DB setup boilerplate across 8 server functions. The approach mirrors the existing `openBaoMiddleware` pattern and is functionally correct.

**Issues Found:** None above reporting threshold.

**Suggestions:**
- [`src/middleware/__tests__/database-middleware.test.ts`] The `nextFn` in the 'injects pool' test is a hand-rolled plain function with a manual counter variable. The existing `openbao-middleware.test.ts` uses `mock()` from `bun:test` for the equivalent, enabling `toHaveBeenCalledTimes(1)` assertions via the standard bun mock API — small consistency improvement to adopt the same pattern here.
- [`src/middleware/database-middleware.ts`] Unlike `openBaoMiddleware` (which caches its client at module scope with `let cachedClient`), this middleware calls `loadDatabaseConfig()` + `databaseConnectionManager.getClient()` on every request. The connection manager is a singleton that caches by host+db key so no pool is leaked, but the extra per-request work is unnecessary. Not a regression from the old per-function code, just worth noting.

**Looks Good:**
- Dynamic imports for `database-client` and `database-config` correctly placed inside `.server()` callback — server-only modules stay out of the client bundle (Critical Rule #4).
- `databaseConnectionManager.getClient()` reuses the pooled singleton correctly — no pool leak.
- Test uses `spyOn` + `mockResolvedValue` + `try/finally { mockRestore() }` — no test pollution.
- All 8 refactored functions follow a uniform pattern; boilerplate reduction is significant.
- Correct exclusion of `stacks/functions.tsx` and `hosts/functions.tsx` per their different patterns.
- No drive-by `console.log` statements.

PR #128, opened 2026-04-19
fix(db): verify TLS cert by default (#103)

jaredglaser:
## Summary

Closes #103.

`DatabaseClient` previously hard-coded `{ rejectUnauthorized: false }` whenever `ssl: true`, silently disabling TLS cert-chain validation and exposing the PostgreSQL connection to MITM attacks.

## Changes

- **`src/lib/clients/database-client.ts`** — `DatabaseConfig` gains optional `sslRejectUnauthorized` field. When building the pool's `ssl` option, use `config.sslRejectUnauthorized ?? true` so cert verification is on by default.
- **`src/lib/config/database-config.ts`** — `loadDatabaseConfig()` now reads `POSTGRES_SSL_REJECT_UNAUTHORIZED`, parsed via a helper that accepts `true`/`false`/`1`/`0` (case-insensitive). Defaults to `true`. Added to the Zod schema.
- **`.env.example`** — documents the new var with a clear MITM warning for self-signed-cert operators.
- **Tests** —
  - New `src/lib/config/__tests__/database-config.test.ts` covering env-unset, `"false"`, `"FALSE"`, `"true"`, `"1"`, `"0"`, and unrecognised-value fallback.
  - Updated `database-client.test.ts` to assert `rejectUnauthorized: true` by default and verified the new field flows through correctly.

## Non-goals

- No change to how `ssl` itself is enabled.
- No CA cert loading — operators with self-signed certs must set `POSTGRES_SSL_REJECT_UNAUTHORIZED=false` until proper CA plumbing is added.

## Test plan

- [x] `bun run typecheck` passes
- [x] `bun test src/lib/config/__tests__/database-config.test.ts src/lib/clients/__tests__/database-client.test.ts` → 29 pass / 0 fail
- [ ] Full `bun test` run in CI (baseline has pre-existing failures unrelated to this change)

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,6 +7,10 @@ POSTGRES_DB="homelab"
 POSTGRES_USER="homelab"
 POSTGRES_PASSWORD="changeme"
 POSTGRES_SSL="false"
+# Verify the PostgreSQL server's TLS certificate when POSTGRES_SSL is enabled.
+# Defaults to "true". Set to "false" ONLY for self-signed certificates — this
+# disables chain validation and exposes the connection to MITM attacks.
+POSTGRES_SSL_REJECT_UNAUTHORIZED="true"
 POSTGRES_POOL_SIZE="10"
 
 # Web Server
--- a/src/lib/clients/__tests__/database-client.test.ts
+++ b/src/lib/clients/__tests__/database-client.test.ts
@@ -51,6 +51,7 @@ const TEST_CONFIG: DatabaseConfig = {
   database: 'testdb',
   user: 'testuser',
   password: 'testpass',
+  sslRejectUnauthorized: true,
 };
 
 describe('DatabaseClient', () => {
--- a/src/lib/clients/__tests__/database-client.test.ts
+++ b/src/lib/clients/__tests__/database-client.test.ts
@@ -63,12 +63,27 @@ describe('DatabaseClient', () => {
     expect(client.id).toBe('postgres://testuser@localhost:5432/testdb');
   });
 
-  it('enables SSL when configured', () => {
+  it('enables SSL with cert verification on by default', () => {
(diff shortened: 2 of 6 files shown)


jaredglaser:
## PR Review

**Summary:** This PR correctly fixes the hardcoded `rejectUnauthorized: false` security bug, replacing it with a configurable field that defaults to `true`. Logic, env var parsing, and test coverage are sound.

**Issues Found:**
- [`src/lib/config/database-config.ts`:11] **Type inconsistency:** `DatabaseConfig` interface declares `sslRejectUnauthorized?: boolean` (optional) but the Zod schema has `z.boolean()` (required). These are structurally compatible so TypeScript doesn't catch it, but any code that validates an externally-constructed `DatabaseConfig` omitting the field would get a Zod runtime error even though the interface permits it. Recommend making the interface field required (`sslRejectUnauthorized: boolean`) since `loadDatabaseConfig()` always resolves it to a boolean — align the interface to match what the schema actually produces.

**Suggestions:**
- [`src/lib/config/__tests__/database-config.test.ts`:30] The 'returns defaults' test doesn't assert `sslRejectUnauthorized`. The nested describe block covers it, but adding `expect(config.sslRejectUnauthorized).toBe(true)` to the defaults test keeps all default assertions in one place.

**Looks Good:**
- Core fix: `config.sslRejectUnauthorized ?? true` correctly defaults to secure cert verification without requiring opt-in.
- `parseBoolEnv` is robust — handles undefined, empty string, case-insensitive true/false/1/0, and falls back to the safe default.
- `.env.example` documents the new var with a clear MITM-attack warning.
- New `database-config.test.ts` is correctly placed in `__tests__/` with `bun:test` imports and good coverage of all parse variants.
- No drive-by console.log, no ASCII banners, no hardcoded colors introduced.

PR #129, opened 2026-04-19
refactor(openbao): unify dual client caches (#107)

jaredglaser:
## Summary

Two independent module-scoped OpenBao client caches existed and were unaware of each other — invalidation in one didn't touch the other, and `initializeOpenBao()` ran per cache:

- `src/middleware/openbao-middleware.ts` — `let cachedClient: OpenBaoClient | null = null`
- `src/data/stacks/functions.tsx` — `let cachedBaoClient: OpenBaoClient | null = null`

This PR extracts a shared `getOpenBaoClient()` factory in `src/lib/clients/openbao-client-factory.ts` that owns the single cache and init guard. Both call sites now delegate to it:

- `openBaoMiddleware` becomes a thin wrapper that calls the factory and stuffs the client into context.
- `stacks/functions.tsx` keeps its local `getOpenBaoClient()` helper as a dynamic-import wrapper around the factory, preserving client-bundle hygiene for this barrel file (per CLAUDE.md rule #4).

### Why factory, not middleware-only?

The middleware-only consolidation would require adding `openBaoMiddleware` to 6 server functions in `stacks/functions.tsx` and changing their call shapes. The factory approach is less invasive, keeps existing call sites identical, and gives a single source of truth for both patterns (middleware-injected clients and direct server-function imports).

### Verification

- `bun run typecheck` clean.
- All 45 openbao-related tests pass (middleware + factory + stacks functions + openbao-server-functions).
- Full suite: same 2146 pass / 9 fail as base — the 9 failures are pre-existing test-isolation flakes in unrelated files (DualSeriesChart, stacks-context) that only fail in the full suite and pass in isolation.
- Grep confirms `cachedBaoClient`/`cachedClient` now live only inside the single factory module.

### Non-goals

- No OpenBao client behavior changes.
- No new cache invalidation features.
- Does not touch the DB-setup region in `stacks/functions.tsx` (separate PR #106).

Closes #107

## Test plan

- [ ] CI: typecheck, tests, coverage pass
- [ ] Sanity-run stacks UI: save compose file, list/get/set/delete variables, ensure OpenBao paths still work end-to-end
- [ ] Confirm server startup still enables the KV v2 engine on first request


jaredglaser:
## PR Review

**Summary:** Centralizing OpenBao client construction into `openbao-middleware.ts` and `openbao-init.ts` is a solid improvement, but the host-management path (`data/hosts/functions.tsx`) is missing the promised cache unification, and the PR introduces unreachable dead code with a static import violation.

**Issues Found:**
- [`src/data/hosts/functions.tsx`:21-25] `loadBaoClient()` creates a fresh `OpenBaoClient` on every invocation with no module-level cache. The PR says "unify dual client caches" but this helper has no cache at all. Both `data/stacks/functions.tsx` (module-level `cachedBaoClient`) and `openbao-middleware.ts` (module-level `cachedClient`) are singletons — this path is not unified. Add a `let cachedBaoClient: OpenBaoClient | null = null` guard here to match.
- [`src/data/hosts/functions.tsx`:48] `addHost` calls `baoClient.ensureSecretsEngine()` directly, bypassing `initializeOpenBao()`. Every sibling handler in the same file (`registerExistingHost` line 69, `verifyHost` line 85) correctly uses `initializeOpenBao(baoClient)`, which provides the promise-singleton retry guard. This inconsistency silently loses the "reset on failure, retry next call" behavior.
- [`src/lib/services/secret-resolver-factory.ts`] `createSecretResolver` is dead code — never imported by any production file (only its own test). It also has a static top-level `import { OpenBaoClient }`, violating the rule that server-only modules must use `await import()` inside handlers. Either wire it into a production call site or remove it.

**Suggestions:**
- `src/lib/services/secret-resolver.ts` and `src/lib/deploy/types.ts` both export a `SecretResolver` interface with different signatures (`resolveSecrets(stack)` vs `resolve(stack, variables)`). These serve different pipeline paths, which is correct — a brief comment on each noting the distinction would prevent future confusion.

**Looks Good:**
- `openbao-middleware.ts` caching with the `initializeOpenBao` promise-singleton is well-structured — correctly handles concurrent first requests and resets on failure for retry.
- Dynamic imports inside the middleware callback follow the server-only module rule and prevent client bundle leaks.
- `OpenBaoClient` constructor injection of `fetchFn` makes the HTTP layer cleanly testable without `mock.module()`.
- Test coverage for the middleware, init service, and secret resolver is thorough with proper reset helpers for test isolation.

PR #130, opened 2026-04-19
refactor(sse): decompose useTimeSeriesStream into composable sub-hooks (#111)

jaredglaser:
## Summary

Closes #111.

`useTimeSeriesStream` previously mixed seven distinct responsibilities across a 365-line, 5-effect, 4-state hook. This refactor decomposes it into composable sub-hooks under `src/hooks/timeSeriesStream/`, leaving the original hook as a thin orchestrator that preserves its **public API and observable behavior exactly**.

### Sub-hooks extracted

- **`useSSEBuffer`** — owns the sorted buffer state, dedup set, pending queue, periodic flush with eviction, and atomic `replaceBuffer` API. Single source of truth for `rows` / `hasData` / `lastDataTimeRef`.
- **`useVisibilityRefresh`** — re-triggers a refresh callback when the document becomes visible, honouring a shared cooldown ref so preload / periodic / visibility refreshes all observe the same guard.
- **`useLatestByEntity`** — memoized latest-per-entity `Map` with **structural sharing** (returns the previous Map reference when no entity's latest row has changed by key).
- **`timeWindowTrim`** (pure helpers, not a hook) — `lowerBound` binary search + `mergeWithEviction`, independently testable without a renderer.

The orchestrator retains:
- One-time preload + `initialData` seeding + stale-initial-data refresh
- Periodic refresh interval
- Preload / SSE / service error composition
- Stale detection interval

### Behavior-preserving judgment calls

- `replaceBuffer` takes an explicit `markHasData` option. The orchestrator passes `true` only on preload seeding; the refresh path does **not** flip `hasData`, matching the original semantics where `hasData` was driven strictly by seed/flush and never by refresh alone.
- `preserveLiveTail` is a separate option so the refresh path can stitch newer in-flight SSE rows onto the replacement buffer exactly as before.
- Getter refs (`getKeyRef`/`getTimeRef`/`getEntityRef`) are passed into sub-hooks rather than the getter functions themselves, preserving the "inline arrow callers don't bust memoization / interval stability" contract.

### Tests

- New per-sub-hook tests in `src/hooks/__tests__/timeSeriesStream/` (30 tests across 4 files) exercise each responsibility in isolation.
- The existing `useTimeSeriesStream` integration suite (18 tests) is the regression safety net and passes unchanged.
- Two new orchestrator tests cover the `preloadFn`-change-after-mount refresh path and the `debug` logging branch, nudging line coverage back over the 99% threshold (baseline on `main` was also 0.02% under).

### Test plan

- [x] `bun test` passes (modulo the pre-existing `DualSeriesChart` / `stacks-context` flakes documented in the task brief — they pass in isolation)
- [x] `bun run test` coverage gate passes: Functions 96.06% (>=95%), Lines 99.01% (>=99%)
- [ ] Manual sanity check of a chart page streaming data (not performed in this isolated worktree)

## Non-goals

No new features, no performance tuning, no SSE endpoint / server route changes, no DataTable or downstream component changes.


jaredglaser:
## PR Review

**Summary:** Cleanly decomposes `useTimeSeriesStream` from a 365-line monolith into four focused units (`useSSEBuffer`, `useVisibilityRefresh`, `useLatestByEntity`, `timeWindowTrim`). Public API is preserved and test surface meaningfully expanded. Two issues to address before merge.

**Issues Found:**

- [`src/hooks/__tests__/timeSeriesStream/useSSEBuffer.test.ts`] **Real `setTimeout` delays in tests.** Every async test case uses `await new Promise(r => setTimeout(r, 60))` to wait for the flush interval. CLAUDE.md requires spying on `globalThis.setTimeout` in `beforeEach`/`afterEach` so timers fire immediately rather than burning real wall-clock time. The existing `useTimeSeriesStream.test.ts` mocks `setTimeout` — these new tests should follow the same pattern.
- [`src/hooks/timeSeriesStream/useSSEBuffer.ts`, `useVisibilityRefresh.ts`] **React refs in `useEffect` dep arrays.** The flush interval in `useSSEBuffer` lists `getKeyRef` and `getTimeRef` as deps; `useVisibilityRefresh`'s effect lists `lastRefreshRef`; the stale-detection effect in `useTimeSeriesStream` gained `lastDataTimeRef`. Refs from `useRef` have stable identity by design — including them in dep arrays is misleading (implies they drive re-subscription) and will likely trigger `react-hooks/exhaustive-deps` warnings. They can be removed safely from all three dep arrays.

**Suggestions:**

- [`src/hooks/useTimeSeriesStream.ts`] The refresh debug log was simplified from `${sorted.length} bucketed + ${liveTail.length} live = ${merged.length} total` to `${rows.length} rows`, losing the live-tail breakdown. Consider emitting the merged count from `replaceBuffer` or logging inside `useSSEBuffer`.

**Looks Good:**

- Sub-hook interfaces are clean and single-responsibility. `useSSEBuffer` correctly owns the dedup set, sorted buffer, pending queue, and flush interval as a unit.
- `useLatestByEntity` preserves the structural-sharing optimization (Map reference equality check before allocating a new Map).
- `timeWindowTrim` extracts `lowerBound` and `mergeWithEviction` as pure functions — straightforwardly testable without a React harness.
- Test file placement follows project conventions; no `mock.module()` on barrel files; no drive-by `console.log` in production paths; no styling regressions.

jaredglaser:
@coderabbitai review

PR #132, opened 2026-04-19
perf(charts): gate rAF loops with IntersectionObserver (#109)

jaredglaser:
## Summary

Perpetual `requestAnimationFrame` loops in time-scrolling charts burn CPU/GPU even when the charts are scrolled off-screen. With N expanded Docker containers, every chart plus its sparkline keeps the main thread churning at 60fps whether anyone is looking at them or not.

This change introduces `useVisibleRAF(targetRef, cb)` — a small hook that uses `IntersectionObserver` to start the rAF loop only when the target element enters the viewport and cancel it the moment it leaves.

Applied to the two continuous-animation sites:

- `useEChartTimeScroll` — the time-axis scroller shared by `ContainerMetricChart`, `DualSeriesChart`, and `ZFSPoolSpeedChart`. The hook now takes a third arg: a ref to the chart's wrapper element. Each caller already had a wrapper `<div>`, so no DOM changes beyond adding a ref.
- `SparklineCanvas` — the per-row canvas sparkline used in the Docker table. The render body moved from a self-enqueuing rAF closure into the `useVisibleRAF` callback, with a few state variables (`lastFrameTime`, `lastDrawTime`) promoted to refs so they survive between frames.

**rootMargin:** `100px`. Charts start animating just before scrolling into view so there is no visible pop-in of a frozen axis when the row reaches the viewport edge. Well-off-screen charts stay idle.

## Falls back cleanly

If `IntersectionObserver` is unavailable (unusual test shims, ancient browsers), the hook falls back to an unconditional rAF loop — behavior identical to before.

## Tests

- `src/hooks/__tests__/useVisibleRAF.test.ts` (new, 8 cases): gated start/stop transitions, stale-closure freedom, null-target no-op, unmount disconnect, IO-absent fallback.
- `src/hooks/__tests__/useEChartTimeScroll.test.ts`: updated to pass a target ref and mock `IntersectionObserver` (immediate-visible).
- `src/components/shared-table/__tests__/SparklineCanvas.test.tsx`: added the same IO mock — existing assertions (rAF called on mount, paths drawn, gradient rebuilt on color change, etc.) all continue to pass unchanged.

`bun run typecheck` clean. Full suite: 2149 pass / 9 pre-existing fails (identical on main — `mock.module()` pollution across test files, independent of this change).

## Test plan

- [x] `bun run typecheck`
- [x] `bun test` (same failure set as main)
- [ ] Manual: expand several Docker containers, scroll them off-screen, confirm CPU drop in DevTools Performance


jaredglaser:
## PR Review

**Summary:** Introduces `useVisibleRAF` to gate rAF loops with `IntersectionObserver`, correctly pausing animation on off-screen charts. The hook implementation is solid with proper cleanup and stale-closure prevention. However, the `SparklineCanvas` test suite has a critical gap introduced by this refactor.

**Issues Found:**

**Critical**
- [`src/components/shared-table/__tests__/SparklineCanvas.test.tsx`:160, 170-178] `SparklineCanvas` now delegates its rAF loop entirely to `useVisibleRAF`, which only calls `start()` after receiving an `isIntersecting: true` callback from `IntersectionObserver`. Happy-DOM ships `IntersectionObserver` but its `observe()` method is a no-op stub — it never fires callbacks. This means rAF is never scheduled in the test environment. At least two tests are broken:
  - "calls requestAnimationFrame on mount" — `rafSpy` call count will be 0
  - "handles empty data array without drawing paths" — `flushOneFrame()` has nothing to flush, so `mockCtx.clearRect` is never called

  Fix: add the same `IntersectionObserver` mock used in `useEChartTimeScroll.test.ts` (fires the callback immediately with `isIntersecting: true` inside `observe()`) to `SparklineCanvas.test.tsx`'s `beforeEach`/`afterEach` blocks. `useVisibleRAF.test.ts` has the exact mock pattern to replicate.

**Suggestions**
- [`src/hooks/useEChartTimeScroll.ts`:19] The `useCallback` wrapping `tick` is redundant — `useVisibleRAF` captures the latest callback via `callbackRef.current = callback` at render time, so the identity of `tick` never matters. Can be a plain inline function or local const.

**Looks Good**
- `useVisibleRAF` cleanup is correct: `observer.disconnect()` + `stop()` (`cancelAnimationFrame`) on unmount — no memory leaks.
- `callbackRef.current = callback` outside the effect is the right pattern to keep the callback fresh without restarting the observer.
- `rootMargin: '100px'` pre-warms animations before elements enter the viewport — good UX choice preventing pop-in.
- `useVisibleRAF.test.ts` is comprehensive: visibility transitions, stale-closure safety, unmount cleanup, null-ref no-op, and `IntersectionObserver` fallback path.
- `useEChartTimeScroll.test.ts` correctly adds the immediately-firing `IntersectionObserver` mock.
- `DualSeriesChart`, `ContainerMetricChart`, and `ZFSPoolSpeedChart` all pass `wrapperRef` on the container div — correct observer target.

PR #147, opened 2026-04-21
feat(stacks): detect and resolve drift between repo and agent (#95)

### LINKED ISSUE #95: Detect and resolve drift between DB and agent stack state
## Problem

The server database and the agent's filesystem can get out of sync. For example:
- Agent container is recreated without a persistent volume → stacks directory is wiped, but DB still shows stacks as deployed
- A deploy succeeds on the agent but the DB update fails → agent has the stack, DB doesn't know about it
- Manual changes on the agent host (someone runs `docker compose down` directly)

This drift causes confusing errors like `404: Stack 'x' not found` on teardown, or the UI showing stale deployment status.

### 1. Drift Detection
On a periodic basis (or on-demand), compare the server DB's view of deployed stacks per host against what the agent actually has:
- **DB says deployed, agent doesn't have it** → "ghost" deployment
- **Agent has it, DB doesn't know** → "untracked" stack
- **Both have it but content differs** → "content drift" (compose file mismatch)

### 2. User Notification
Surface drift in the UI (e.g., a warning badge on the Stacks page or per-host status). Show a summary like:
- 2 stacks in DB not found on agent
- 1 stack on agent not tracked in DB

### 3. Drift Resolution
Let the user review each drift item and choose which source of truth to use:
- **Trust DB** → re-deploy the stack to the agent (push DB state to agent)
- **Trust Agent** → update DB records to match agent state (pull agent state to DB)
- **Remove** → tear down on both sides

The user should be able to resolve items individually or in bulk.

## Context

This came up during local development where agent container restarts wiped the stacks directory (now fixed with a named volume in docker-compose). But the general problem applies to production too — network failures, partial deploys, and manual interventions can all cause drift.

jaredglaser:
## Summary

Adds drift detection between the repo manifest and each Docker host's agent filesystem, plus a resolution flow to trust the repo, trust the agent, or remove on both sides. Closes #95.

- Agent `GET /stacks/inventory` returns per-stack sha256 compose hashes; `GET /stacks/compose?stack=name` returns a single compose file so the server can adopt agent state.
- Classifies items as `ghost` (in repo, not on agent), `untracked` (on agent, not in repo), or `content` (both, hashes differ).
- Re-scans before acting so a stale UI click cannot drive a destructive resolution.
- If the target host errored during the re-scan, surfaces the real message instead of "no drift item found".

## Type safety

- `StackDriftItem` is discriminated on `kind` so consumers cannot read a hash that is not present for that variant.
- `StackDriftResolutionResult` is discriminated on `outcome` so the UI does not non-null-assert `deployId` / `commitSha`.
- `resolveDriftSchema` resolution enum is locked against `StackDriftResolution` via `satisfies`.

## Robustness

- `loadRepoStacks` narrows its catch to `FileNotFoundError`, logs and rethrows everything else. A corrupt git ODB no longer masquerades as "every agent stack is untracked".
- `readAgentCompose` detects 404 via `AgentClientError.statusCode === 404` instead of substring-matching error messages.
- `syncRepoStackToAgentCompose` throws if the target stack already exists in the manifest on a different host (instead of silently rewriting the host assignment). Also drops a spurious `commitSha!` non-null assertion.
- Agent `handleStackInventory` skips individual unreadable compose files (EACCES / EIO / TOCTOU delete / EISDIR) rather than aborting the whole inventory. One bad directory no longer makes the scanner think the host has zero stacks.
- Agent `handleGetStackCompose` catches `readFileSync` failures and returns 500 with a message instead of throwing out of the request handler.
- `scanStackDrift` logs per-host failures server-side in addition to pushing to `scanErrors` for the UI.

## UI

- Drift summary alert on `/stacks` with per-item resolve buttons.
- Drift warning on `/stacks/$stackName` for the stack being viewed.
- Drift marker icon in the side nav for any stack with drift.
- Resolve buttons disabled while a scan or resolve is in flight.
- Drift query kept warm (`staleTime: 60s`, no refetch on focus) so navigating between list and detail does not re-fan-out to every agent.

## Tests

- `stack-drift-service.test.ts`: classification (ghost, untracked, content), scan-error propagation, sort stability, non-docker hosts skipped, all four resolution paths including `teardown-pending` and synchronous `remove`, TOCTOU (item disappears after scan), scan-error during resolve, repo copy disappears before `trust_repo`, agent compose disappears before `trust_agent`.
- `StackDriftSummary.test.tsx`: empty / loading / scan-error / disabled-while-resolving / disabled-while-loading.
- Agent inventory: empty dir, unreadable compose entries skipped, 500 when the stacks dir is not a dir. Compose handler: 400 for invalid stack name, 400 for missing stack query param.
- Existing `agent-client.test.ts`, `functions.test.ts`, `StackNav.test.tsx` extended for the new endpoints, server functions, and drift marker.

## Test pollution fix

`StackNav.test.tsx` used `mock.module('@/components/stacks/stacks-context', ...)` which is process-scoped under bun and leaks into `stacks-context.test.ts`. That was latent on main and got exposed when the new drift tests reordered the suite. The fix renders real `StackListContext` / `StackStatusContext` providers instead.

## Test plan
- [x] `bun run typecheck`
- [x] `cd agent && bun run typecheck`
- [x] `bun test` (2361 pass, 0 fail)
- [x] `cd agent && bun test` (215 pass, 0 fail)
- [ ] Manual: scan with one agent offline, confirm partial report + scanError
- [ ] Manual: trigger each resolution path (trust_repo / trust_agent / remove) on each drift kind

### DIFF
--- a/agent/src/__tests__/stacks.test.ts
+++ b/agent/src/__tests__/stacks.test.ts
@@ -4,6 +4,8 @@ import {
   handleStackTeardown,
   handleStackRestart,
   handleStackStatus,
+  handleStackInventory,
+  handleGetStackCompose,
   parseContainerNames,
 } from '../routes/stacks';
 import { mkdirSync, rmSync, existsSync, readFileSync } from 'fs';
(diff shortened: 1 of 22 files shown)


jaredglaser:
Superseded by #306.

This branch (`feat/issue-95-drift`) had no common ancestor with current `main` (orphaned history) and was written against a since-changed codebase (MUI removed, `StackNav` replaced by `StacksOverview`, agent start/stop routes added, Ed25519 JWT agent auth). #306 is a clean re-implementation of the same drift detection/resolution feature on top of current `main`, preserving this branch's history untouched. Closing in favor of #306.

---

Issue #232, opened 2026-06-11
TimescaleDB tuning: 1-day chunks, continuous aggregate for history, explicit retention


jaredglaser:
**Problem:** Stats hypertables use the default 7-day `chunk_time_interval`, there are no retention policies (removed in migration 007), and no continuous aggregates: history preloads `time_bucket` over raw 1 Hz rows.
**Fix:** New migration(s): set `chunk_time_interval => INTERVAL '1 day'` for the stats hypertables (affects new chunks only), add a real-time continuous aggregate (1-minute buckets) for the history path and point the history queries at it for long windows, and add an explicit retention policy where raw data drops after N days while the cagg survives.

Source: finding F5/F6 in docs/data-flow-review.md (https://github.com/jaredglaser/homelab-manager/pull/226). Part of epic #227.

PR #276, opened 2026-06-12
feat(db): TimescaleDB chunk tuning and three-tier stats retention cascade

jaredglaser:
## Summary

Now carries two merged pieces: the chunk interval tuning, and the three-tier retention cascade from #370 which merged into this branch.

Implements the retention model the Data Retention settings UI has always described but nothing ever applied: **raw to 1-minute to 1-hour, with the hourly tier kept forever**, across all three stats hypertables.

| Tier | Chunk | Compress after | Retention |
|---|---|---|---|
| raw 1 Hz | 1 hour | 6 hours | 24 hours |
| `_1m` | 1 day | 7 days | 30 days |
| `_1h` | 7 days | 30 days | never |

Migrations added: `029_stats_chunk_tuning.sql`, plus `030`/`031`/`032` for docker, zfs and proxmox.

## Three invariants

**Hourly is built on the minute aggregate, not on raw.** If `_1h` read raw, raw's 24-hour retention would become its recovery window: a refresh job down for two days could never materialize the gap, and a refresh recomputing a region whose raw chunks are gone would delete the aggregate for that region. Sourcing from the 30-day `_1m` gives 30 days of slack.

**Each refresh window sits strictly inside its source's retention.** `_1m` over raw uses `start_offset 6 hours`; `_1h` over `_1m` uses `start_offset 3 days`.

**Weighted averages, not average-of-averages.** `AVG()` over the minute `AVG()`s is only correct when every minute has identical sample counts. `_1m` carries `count(*) AS sample_count`; `_1h` materializes `SUM(metric * sample_count)` and `SUM(sample_count)`; a plain view divides. On a realistic sparse hour this is the difference between `10.27027027027027` and `11.33333`.

## Per-source column treatment

**proxmox is the one that needed care.** `AVG()` on exactly four true gauges (`cpu`, `mem`, `disk`, `storage_avail`), and **`last()` on `netin`, `netout` and `uptime`**. Those arrive from the Proxmox API as cumulative totals with no delta applied by the collector, and `GuestSection.tsx` renders them as raw byte totals. Averaging a monotonic counter yields the bucket midpoint, which is meaningless and looks plausible. Also `last()` on the capacity constants and the identity columns. A test asserts `AVG(netin)` never appears.

docker averages all eight metrics (gauges or pre-computed rates) with `last()` on `container_name`/`image`; zfs averages its seven with `last()` on `entity_type`/`indent`.

## Backfill and retention run as a startup task

The migrations contain only transactional DDL. The initial backfill and the retention policies live in `runStatsRollupBackfill` (`src/lib/database/stats-rollup-backfill.ts`), fired from `initServer()`. Per source: read the real `min(time)`, refresh `_1m` in 2-day windows then `_1h` in 6-day windows, and **only after both succeed** add retention. A failed window logs, leaves that source without retention, and the next boot retries against history that is still there.

That ordering matters because migration `007` removed retention entirely, so a long-running deployment has months of raw stats. Applying retention before a successful rollup would destroy it.

Guarded by `pg_try_advisory_lock` on a key distinct from the migrations lock, so a second instance skips rather than blocking and a long backfill never blocks a migrating process.

### Why not in SQL

`CALL refresh_continuous_aggregate` cannot run inside a transaction block, and node-postgres wraps a multi-statement string in an implicit transaction. Doing it in SQL required a `-- migrate:no-transaction` mode and roughly 70 lines of hand-rolled SQL lexer in `migrate.ts`, which already had two gaps (nested block comments, `E''` backslash escapes). All of it is gone: `migrate.ts` and its test are byte-identical to their pre-cascade state, and each `CALL` is now its own `client.query`.

## Routing takes the coarser of span and age

`resolveStatsTier` computes a tier from window span and from the age of the window start, then takes the maximum. Without the age dimension, a short window pointed at a distant past (a panned detail chart) routed to raw by span, and raw is gone after 24 hours, so the chart rendered empty. Since the dashboard chart is capped at 1800s and cannot reach the aggregates, `getDockerStatsForContainer` is their only consumer, so that bug would have made the tiers unreachable in the one place they exist to serve.

`getContainerInfo` now reads from `docker_stats_1m`, so a container idle more than 24 hours keeps its name and image.

## Settings

The three Data Retention controls are set to the hardcoded values and disabled; the hourly control becomes a static "Kept forever" row. The `retention` plumbing stays for a future PR that makes these operator-configurable.

## Known limitations

- **NULL dilution in the weighted average**: `SUM(metric * sample_count)` skips NULL products while `SUM(sample_count)` still counts those minutes. Docker and zfs fill all columns per row and proxmox NULLs are systematic per `entity_id`, a group key, so this should not bite in practice.
- Windows narrower than one bucket are skipped; `refresh_continuous_aggregate` rejects them and the refresh policy covers that tail.
- The 1800s chart window cap is untouched.

## Testing

`bun run typecheck:all` clean. `src/lib/database/**`: 298 pass, 0 fail.

**No TimescaleDB instance was available, so no migration and no backfill statement was executed.** The SQL is unverified against a live server. #369 adds the CI job that would close this gap; merging it first and rebasing this onto it would let the cascade land on evidence.

Relates to #232.


jaredglaser:
## Triage verdict: Split (confidence: high)

I verified the two things this PR claims about itself and both hold: in migration 025 the backfill really does precede the raw retention policy (CREATE WITH NO DATA at :15, CALL refresh at :39, add_retention_policy on docker_stats last at :58), and every statement in the no-transaction file is idempotent. The problem is that the three pieces here have three different answers, so I would split rather than land or reject as a unit.

The chunk tuning (024:9-11) is non-destructive, affects only chunks created after it runs, and depends on nothing else; that can go in now.

The continuous aggregate currently has no reader: the read path is gated on seconds > 3600 in stats-repository.ts:155, but chartWindowSeconds is clamped to 1800 in settingsAtom.ts:166 and the slider maxes at 1800 in settings.tsx:181, so the largest value that can reach getDockerStatsHistory is 1810. As written docker_stats_1m would be refreshed every 5 minutes, compressed and kept 365 days with nothing reading it. Either raise the chart window cap or point getDockerStatsForContainer at the aggregate first.

The retention half needs your call before anything else: main already ships a Data Retention section in settings.tsx:221-271 with raw hours, minute aggregate days and hour aggregate days sliders, and those keys are pure settings plumbing with nothing applying them to TimescaleDB. This PR hardcodes a different model that ignores all three, so after merge dragging the Raw Data slider still does nothing and the UI would actively misreport the database. Separately zfs_stats and proxmox_stats get 30-day retention with no aggregate, and getZFSStatsHistory, getProxmoxStatsHistory and getContainerInfo (stats-repository.ts:276-296, which resolves a container's name and image from docker_stats) all read the raw tables, so a container idle for 31 days loses its name and image, not just its chart.

Two rebase notes: pool.query('BEGIN') was already at the merge base so this PR did not introduce it, but main has since moved to a dedicated PoolClient with a session advisory lock (migrate.ts:23-36), so take main's side there and expect src/lib/database/__tests__/migrate.test.ts to need a rewrite rather than a conflict resolution, since runMigrations no longer accepts a migrations directory. And main's highest migration is 026, with #278 and #279 also claiming 024, so these need renumbering to 029 and 030.

### Specific items to fix
- docker_stats_1m has no reader: cagg gate is seconds > 3600, chart window is capped at 1800 (max reachable 1810)
- Retention scheme contradicts the shipped Data Retention settings UI on main, which remains unimplemented
- zfs_stats and proxmox_stats get 30-day retention with no aggregate; getZFSStatsHistory, getProxmoxStatsHistory and getContainerInfo all read raw
- Unbounded CALL refresh_continuous_aggregate at 025:39 runs synchronously on boot with no rollback, so a timeout re-runs the whole file next boot
- migrate.test.ts cannot be conflict-resolved; runMigrations on main has no migrationsDir parameter

**Effort:** chunk-tuning split, under an hour. Full PR as scoped, 1 to 2 days, and it should not proceed until the retention product model is settled.

**Obsolescence:** no, but conflicting. Proof: no docker_stats_1m or continuous aggregate exists in main's migrations or src; main's migrate.ts has no no-transaction directive; issue #232 open. However origin/main:src/routes/settings.tsx:221-271 already ships a Data Retention UI (rawDataHours max 168, minuteAggDays max 30, hourAggDays max 365) whose values are never applied to TimescaleDB, so main has a declared retention model this PR does not implement.

- `origin/feat/f5-f6-timescale-tuning:src/lib/database/repositories/stats-repository.ts:155 (if seconds > 3600) vs origin/main:src/hooks/settingsAtom.ts:166 (Math.min(1800, ...)) and src/routes/settings.tsx:181 (max={1800})`
- `origin/main:src/routes/settings.tsx:221-271 (Data Retention sliders); rawDataHours appears only in settings plumbing`
- `migrations/024_timescale_chunk_retention.sql:9-11 (set_chunk_time_interval) and :21-22 (zfs/proxmox 30-day retention, no aggregate)`
- `origin/main:src/lib/database/repositories/stats-repository.ts:276-296 (getContainerInfo reads raw docker_stats), :300-322 (zfs raw), :415-421 (proxmox raw)`
- `git show 5d66d9f:src/lib/database/migrate.ts:52 -> pool.query('BEGIN') already at merge-base`
- `origin/main:src/lib/database/migrate.ts:23-36 (advisory lock, PoolClient, no migrationsDir parameter)`
- `migrations/025_docker_stats_cagg.sql:39 (unbounded CALL refresh_continuous_aggregate on the boot path, no rollback)`
- `origin/main:migrations/007_remove_retention_policies.sql (prior decision to drop retention)`
- `git ls-tree origin/main migrations/ -> highest is 026_container_events_ports_mounts.sql`

</details>

---

PR #278, opened 2026-06-12
feat(deploy): queue git pushes blocked by an active deploy

### LINKED ISSUE #241: Queue git pushes that arrive during an active deploy
**Problem:** The partial unique index (`migrations/011_deploy_active_unique.sql`) correctly enforces one active deploy per stack+host, but the pipeline reports a blocked request as failed and nothing retries it. A git push landing while a deploy runs means the newest commit silently never deploys. Note: `deduplicatePending` (`deploy-repository.ts:145`) is dead code in practice since at most one pending/in_progress row can exist.
**Fix:** On a 23505 rejection for a `git_push` trigger, record the request as superseded/queued and re-run the newest queued request for that stack+host when the active deploy reaches a terminal state (the pipeline already has the completion hook where it fires NOTIFY). Converges to latest-desired-state semantics.

Source: finding F16 in docs/data-flow-review.md (https://github.com/jaredglaser/homelab-manager/pull/226). Part of epic #227.

jaredglaser:
## Summary

The partial unique index (`idx_deploy_one_active_per_stack_host`, migration 011) allows one active deploy per stack+host, but a git push landing while a deploy runs was recorded as failed and never retried, so the newest commit silently never deployed.

- New `deploy_queue` table (migration `024_deploy_queue.sql`): one row per stack+host (PK), stores the full `DeployRequest` as JSONB plus `queued_at`, FK to `managed_hosts` with cascade delete.
- `DeployRepository.enqueueDeploy` upserts so the newest blocked push wins; the conditional upsert (`queued_at <= EXCLUDED.queued_at`) prevents a stale drained request from replacing a newer queued one. `dequeueDeploy` is a DELETE RETURNING.
- `DeployPipeline.execute`: a 23505 rejection for a `git_push` trigger now queues the request and returns a new `queued` result status. Blocked UI deploys (and manual rollbacks) keep the immediate failed response.
- Queue drain runs on the pipeline completion paths where `notifyStackChange` fires: dispatch success, dispatch failure, dispatch throw, postSuccess failure, and `resumePending` secret-resolution failure. The drained request goes through the full `execute` path; if it loses the insert race against a concurrent push it is re-enqueued with its original timestamp.
- `deduplicatePending` removed: the unique index guarantees at most one pending/in_progress row per stack+host, so it was dead code.
- Docs updated (CLAUDE.md, docs/architecture.md).

Migration numbering: this branch and the F5/F6 branch (#232) both claim number 024; whichever merges second gets renumbered on rebase.

## Testing

- `bun run typecheck` clean.
- `bun test --isolate` on the three touched test files: 75 pass, 0 fail.
- New tests: blocked git push queues (and UI deploy still fails immediately), queueing failure falls back to failed, drain on success / failure / dispatch throw, newest-wins re-enqueue with original timestamp on insert race, drain resilience when dequeue or the drained request throws, resumePending failure drains, repository upsert/dequeue SQL shape.
- Touched source files at 100% function and line coverage.

Scope is completion-hook drain only. Follow-ups worth filing: drain on startup recovery and `DeployWatchdog` timeouts (a queued row currently waits for the next deploy on that stack+host if the process crashes mid-deploy), and drain after UI `rejectPendingDeploy` (stack-service path, outside the pipeline).

Closes #241

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -167,7 +167,7 @@ OIDC login, required by default; `AUTH_DISABLED=true` is the explicit opt-out an
 
 ### Deploy Pipeline (`src/lib/deploy/`)
 
-Trigger-agnostic orchestration: `DeployRequest` → validate → resolve secrets → dispatch to agent → record result. Each caller (`src/lib/git/post-receive-handler.ts` for git pushes, `src/lib/stacks/stack-service.ts` for UI deploys/rollbacks) assembles its own `DeployRequest` inline; there is no separate trigger-builder layer. Concurrency enforced via PostgreSQL partial unique index. Stuck deploys recovered on startup and via `DeployWatchdog` (default 20-min threshold) so a crashed process can't leave `in_progress` rows stranded. Details: [Deploy Pipeline](docs/architecture.md#deploy-pipeline-srclibdeploy).
+Trigger-agnostic orchestration: `DeployRequest` → validate → resolve secrets → dispatch to agent → record result. Each caller (`src/lib/git/post-receive-handler.ts` for git pushes, `src/lib/stacks/stack-service.ts` for UI deploys/rollbacks) assembles its own `DeployRequest` inline; there is no separate trigger-builder layer. Concurrency enforced via PostgreSQL partial unique index; a git push blocked by an active deploy is queued in `deploy_queue` (newest per stack+host wins) and dispatched after the active deploy reaches a terminal state, while blocked UI deploys fail immediately. Draining checks `getLatestSuccessful` first and discards the queued commit if a newer one already deployed, so a stranded queue row can't revert a stack. A queued push also gets a `queued` `deploy_history` row so it shows up in the history list and on the stack-status channel. Stuck deploys recovered on startup and via `DeployWatchdog` (default 20-min threshold) so a crashed process can't leave `in_progress` rows stranded. Details: [Deploy Pipeline](docs/architecture.md#deploy-pipeline-srclibdeploy).
 
 ### Git Management (`src/lib/git/`)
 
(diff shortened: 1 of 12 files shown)


jaredglaser:
## Triage verdict: Rebase, then merge (confidence: medium)

The problem is real and this is the right shape for it. I confirmed the design details hold up: managed_hosts.name is UNIQUE (migrations/009:4) so the new FK is valid, dequeue is DELETE RETURNING so it is race-safe, and resumePending does end in dispatch, so the manual-approval path is drained, which the description understates.

Two things should be closed before merge. First, there is no staleness check on a dequeued request, so a queued commit can deploy over a newer one. If the process crashes mid-deploy, startup recovery and DeployWatchdog both mark the active deploy failed without draining, the queued row persists, and the next successful deploy drains it and deploys the older commit, reverting the stack. The same happens without a crash when a drained request loses the insert race, is re-enqueued with its original queued_at, and is drained after the newer commit deployed. The queued_at guard only stops an older row replacing a newer queued row. The cheapest fix is to compare the dequeued request's commitSha against deployRepo.getLatestSuccessful(stack, host) inside drainQueue, a method already used on the change-detection path at pipeline.ts:243.

Second, widening PipelineResult.status to DeployStatus | 'queued' breaks the build against current main in a file this PR does not touch: src/lib/stacks/stack-mappers.ts:93 requires status: DeployStatus and stack-service.ts:147 and :171 pass pipeline.execute straight in. That line was Promise<{ deployId?: number; logs?: string }> at the merge base, so main tightened it after this branch was cut, which is why the 2026-06-12 CI run cannot have caught it. Either widen stack-mappers.ts or keep 'queued' out of the shared type.

Two smaller notes: queueBlockedPush returns before any insertDeploy, so a queued push produces no deploy_history row at all and is invisible to the history list and the stack-status channel, not just missing a badge; and the branch's resolveEnv predates the 'update' action that main added at pipeline.ts:232, so that file must not resolve in favor of the branch. I would also drop the deduplicatePending removal, which is unrelated cleanup that enlarges the conflict surface in two already-conflicted files. Main's highest migration is 026 and both #276 and #279 also add a 024, so please renumber this to 028.

### Specific items to fix
- No staleness check on dequeue: a queued commit can deploy over a newer one and silently revert the stack
- PipelineResult.status widening breaks tsc against main via stack-mappers.ts:93, invisible to git merge-tree and to the stale CI
- Queued pushes produce no deploy_history row, so they are invisible to the history list and the stack-status channel
- resolveEnv on the branch predates the 'update' action; a careless merge gives update deploys an empty .env
- Migration numbered 024, colliding with main's existing 024 and with #276 and #279

**Effort:** half a day: staleness guard plus test, stack-mappers type fix, renumber to 028, drop the deduplicatePending removal, re-run CI.

**Obsolescence:** no. Proof: no deploy_queue table or enqueue path exists in main's src/lib/deploy/ or migrations/; origin/main:src/lib/deploy/pipeline.ts:162-164 still returns failed on a null deployId with no retry path; issue #241 open.

- `origin/main:src/lib/deploy/pipeline.ts:162-164 (returns failed on null deployId) and :243 (getLatestSuccessful available for a staleness guard)`
- `origin/main:src/lib/git/post-receive-handler.ts:114-118 (only console.info)`
- `origin/main:src/lib/stacks/stack-mappers.ts:93 (status: DeployStatus) and stack-service.ts:147,171 (pipeline.execute passed directly)`
- `git show 5d66d9f:src/lib/stacks/stack-mappers.ts:93 -> Promise<{ deployId?: number; logs?: string }> at merge-base`
- `origin/feat/f16-deploy-queueing:src/lib/deploy/pipeline.ts:10 (status: DeployStatus | 'queued'), :217-221 (queueBlockedPush returns before insertDeploy)`
- `origin/main:src/lib/deploy/pipeline.ts:232 (resolveEnv includes the 'update' action; branch does not)`
- `origin/main:src/lib/database/repositories/deploy-repository.ts:160 (deduplicatePending still live)`
- `git merge-tree -> 5 conflicts; migrations/024_deploy_queue.sql collides with main's 024_deploy_started_at.sql and with #276/#279`

</details>

---

PR #279, opened 2026-06-12
perf(git): replace per-request token decryption with indexed hash lookup

### LINKED ISSUE #243: Git HTTP auth: O(n) token decryption per request
**Problem:** `src/routes/api/git.$.ts` loads all `git_tokens` rows and JWE-decrypts each on every git HTTP request before the timing-safe compare.
**Fix:** Store `SHA-256(token)` in an indexed column at creation (the sessions table already uses exactly this pattern) and look up O(1). Migration needed for existing tokens (re-issue or backfill on next use).

Source: finding F18 in docs/data-flow-review.md (https://github.com/jaredglaser/homelab-manager/pull/226). Part of epic #227.

jaredglaser:
## Summary

Git HTTP auth (`src/routes/api/git.$.ts`) loaded every `git_tokens` row and JWE-decrypted each one on every request before the timing-safe compare, making auth O(n) in token count with a JWE decrypt per row.

- Migration `027_git_token_hash.sql` adds a `token_hash` column (SHA-256 hex of the raw token) with a unique index. The column stays nullable; `NOT NULL` is a follow-up once the backfill has been watched against real data.
- `createGitToken` writes the hash at creation (`hashGitToken` in `src/lib/git/git-token-auth.ts`).
- Auth is a single indexed lookup by `token_hash`, returning null on a miss. There is no decrypt-and-compare fallback.

### Why a startup backfill instead of a fallback

`token_hash` covers the raw token, but the database only holds `encrypted_token` (JWE). A SQL migration cannot compute the hash, because decryption needs the master key and that lives in the app environment.

An earlier revision of this PR handled that with a lazy fallback in the auth path. That had two costs:

- Any failed auth, including an unauthenticated probe, fell through and JWE-decrypted every NULL-hash row. Cheap request, expensive server work.
- It never converged. A token that is never presented again is never hashed, so the scan never goes away.

`backfillGitTokenHashes` now runs once at startup from `src/lib/server-init.ts`, alongside `startDeployRecovery`. It selects `WHERE token_hash IS NULL`, decrypts each row with the keyring, and writes the hash.

Failure handling:

- The backfill never throws, so it cannot take down startup.
- A row that fails to decrypt (key rotated away, corruption) is logged at `console.error` and skipped; the loop continues and the row keeps a NULL hash for the next start to retry.
- If the keyring itself is unavailable, every row fails and an additional `console.error` says so explicitly. Nothing is written, and the next start retries.

### Why deleting the fallback is safe

The fallback's only purpose was handling rows that CAN be decrypted, and the backfill handles exactly those. A row whose decrypt fails cannot authenticate on `main` today either, because `main` decrypts in order to compare. Those rows lose nothing.

### Side effect: git auth no longer needs the master key

With the fallback gone, `src/routes/api/git.$.ts` imports no crypto at all: `loadMasterKeyring`, `decryptValue`, and the lazy `loadCrypto` memoization are removed. Git HTTP auth keeps working when `MASTER_KEY` is unavailable, once the backfill has run.

### Other changes

- Dropped the redundant `timingSafeEqual` on the lookup result. The row is selected by `WHERE token_hash = $1`, so the comparison could only ever return true; `findByTokenHash` no longer selects `token_hash` either.
- Repository: `findAllEncrypted` (sole caller was the route) became `findMissingHash`, now feeding the backfill rather than auth; added `findByTokenHash` and `setTokenHash`.

Migration numbering note: open PRs #276 and #278 also claimed 024, hence 027.

## Testing

- `bun run typecheck:all` clean (homelab-manager + agent).
- `bun test --isolate src/lib/git/__tests__/git-token-auth.test.ts` (9 pass): indexed lookup match, unknown token returns null after one lookup with no crypto, backfill hashes every NULL row, backfill skips a row that fails to decrypt and continues, a failed hash write counts as unhashed without aborting, keyring-unavailable logs loudly and leaves all rows intact, a backfilled token then authenticates on the indexed path.
- `bun test --isolate src/lib/database/repositories/__tests__/git-token-repository.test.ts` (17 pass): create inserts `token_hash`, `findByTokenHash`, `findMissingHash` filters NULL-hash rows, `setTokenHash`.
- `bun test --isolate src/data/__tests__/git-tokens.functions.test.ts` (8 pass): created token stores the SHA-256 hash of the raw token.

Closes #243


jaredglaser:

This is the cleanest PR in the open set and it merges with zero conflicts against main today. On the security design, unsalted SHA-256 is the right call and I would not change it: the tokens are randomBytes(32) hex, so 256 bits of entropy, and salting exists to defeat precomputation against a low-entropy input space, which does not apply. It also matches src/lib/auth/session-manager.ts:22 and :45, which is the precedent issue #243 points at. Backfill-on-match is safe too, since the hash is only written after a successful compare against the decrypted plaintext, so a caller who does not already hold the token cannot poison it, and I confirmed key rotation is unaffected because scripts/migrate-master-key.ts:229-235 re-encrypts encrypted_token only and does not change the plaintext. The test file is genuinely good: git-token-auth.test.ts asserts findLegacyEncrypted and decryptToken are not called on the fast path (:51-52), covers a malformed stored hash falling through without throwing, per-row decrypt failure, and backfill write failure not throwing.

One thing to fix before merge: moving loadMasterKeyring into the per-token decryptToken closure means it now runs once per legacy row, and it is not memoized (src/lib/crypto/master-key.ts:29-52 re-reads env and MASTER_KEY_FILE and re-imports every key on each call). That is bounded, since each token is backfilled on its first successful auth, but the unauthenticated path still forces a full scan with garbage credentials and lazy backfill never converges for tokens that are never used again. Hoisting the keyring into a lazily initialized promise outside the closure keeps the never-touch-the-keyring-on-the-fast-path property, and a one-time startup backfill would close the window entirely.

Two smaller notes: the timingSafeEqual on the fast path compares a value selected by WHERE token_hash = $1 against itself, so it can only return true and the comment overstates it; and one threat-model shift worth stating, that after this change a database write is sufficient to mint a working git token, where before it required a valid JWE and therefore the master key. Sessions already have that property so it is not a practical downgrade, but it is being traded away deliberately. Finally, main's highest migration is now 026 and both #276 and #278 also add a 024, so please renumber this to 027.

### Specific items to fix
- loadMasterKeyring moved inside the per-row decrypt closure; not memoized, so N legacy rows means N file reads and N key imports per request until backfill converges
- Lazy backfill never converges for tokens never used again, so an unauthenticated caller can force the full decrypt scan indefinitely
- Migration numbered 024, colliding with main's existing 024_deploy_started_at.sql and with #276 and #278

**Effort:** 1 to 2 hours: hoist the keyring, renumber the migration to 027, optionally add a startup backfill.

**Obsolescence:** no. Proof: origin/main:src/routes/api/git.$.ts:80-105 still hoists the keyring and runs the full decrypt loop; no git_tokens.token_hash exists in main's migrations; issue #243 open. findAllEncrypted has exactly one production caller (git.$.ts:81), so the rename to findLegacyEncrypted is safe.

- `origin/main:src/routes/api/git.$.ts:80-105 (hoisted keyring plus full decrypt loop, still present)`
- `origin/perf/f18-git-token-hash-lookup:src/routes/api/git.$.ts:73-78 (loadMasterKeyring inside decryptToken)`
- `origin/main:src/lib/crypto/master-key.ts:29-52 (not memoized: collectKeyEntries plus importRawKey per call)`
- `origin/main:src/lib/auth/session-manager.ts:22,45 (identical unsalted SHA-256 precedent)`
- `src/data/git-tokens.functions.tsx (randomBytes(32).toString('hex'))`
- `src/lib/git/__tests__/git-token-auth.test.ts:51-52 (asserts findLegacyEncrypted and decryptToken are not called on the fast path)`
- `git merge-tree merge-base origin/main branch | grep -c '<<<<<<<' -> 0`
- `scripts/migrate-master-key.ts:229-235 (re-encryption touches encrypted_token only, so token_hash survives key rotation)`
- `git ls-tree origin/main migrations/ -> highest is 026_container_events_ports_mounts.sql`

</details>

---

PR #281, opened 2026-06-12
fix(auth): add Max-Age and __Host- prefix to session cookie

### LINKED ISSUE #247: Session cookie: add Max-Age and __Host- prefix
**Problem:** `buildSessionCookie` (`src/lib/auth/callback-handler.ts:38-41`) sets no `Max-Age`/`Expires` (browser-session cookie backed only by the DB TTL) and no `__Host-` prefix.
**Fix:** Set `Max-Age` to `SESSION_TTL_HOURS * 3600` and use the `__Host-` prefix when `Secure` applies (requires reading the cookie under both names during a transition, or accepting one forced re-login).

Source: finding F22 in docs/data-flow-review.md (https://github.com/jaredglaser/homelab-manager/pull/226). Part of epic #227.

jaredglaser:
## Summary

- The session cookie was a browser-session cookie with no Max-Age, so it expired with the browser while the DB session stayed valid for SESSION_TTL_HOURS. It also lacked the __Host- prefix on https deployments.
- New shared module `src/lib/auth/session-cookie.ts` owns the cookie name, build, clear, and parse logic. `buildSessionCookie` now sets `Max-Age` from `SESSION_TTL_HOURS` (hours * 3600).
- A deployment uses exactly one cookie name, keyed on isSecure: https issues `__Host-session` (browser enforces Secure, Path=/, no Domain), http (local dev) keeps `session` since browsers reject the prefix without Secure.
- Reads resolve the active name in one place. `resolveUserFromCookie` now reads through `parseSessionCookie(isSecureCookie())` instead of the hardcoded plain name, so the auth middleware, SSE auth, and the session-read server fn in `auth.functions.tsx` all follow from that single seam. `isSecureCookie()` derives isSecure from `OIDC_REDIRECT_URI` and stays separate from `loadAuthConfig()`, which would throw on an unauthenticated request when the OIDC env vars are incomplete. Logout reuses the isSecure it already derives from config. On https a plain `session` cookie is ignored, so a cookie planted under the plain name cannot fixate a session past the __Host- guard.
- `parseCookie` moves from `resolve-user.ts` into `session-cookie.ts` so the auth path has one cookie parser rather than two copies of the same regex.
- The `AUTH_DISABLED` logout path built its clear-cookie from a hardcoded `false`, which left a stale `__Host-session` cookie behind on https deployments. It now derives the name from `isSecureCookie()` like every other site.
- `CallbackHandlerDeps` gains `sessionTtlHours`, wired from `loadAuthConfig()` in `callbackGetHandler`.

## Migration

There is no dual-name support: the cookie is never read or cleared under two names at once. On https deployments this means every live session is invalidated on deploy, since the browser still holds `session` while the server now reads only `__Host-session`. Each user is bounced to `/login` once and re-authenticates through the OIDC provider. No data migration or operator action is required, and http deployments are unaffected because the name does not change there.

## Testing

- `bun run typecheck:all` passes.
- Co-located test file `src/lib/auth/__tests__/session-cookie.test.ts` covers name selection, Max-Age, Secure/__Host- attributes, single-name clearing, and parse precedence/decoding edge cases (including that the plain name is ignored on https, the __Host- name is ignored on http, and that `__Host-session=` does not satisfy a read for the plain `session` name).
- `resolve-user.test.ts` covers the secure-name read path and asserts a plain `session` cookie is rejected on an https deployment without ever building a session manager.
- Callback, logout, and auth.functions tests updated for the new deps shape and single-name reads. Note: Happy-DOM hides Set-Cookie on Response objects, so attribute assertions live in the helper tests, matching the existing test pattern.

Closes #247

### DIFF
--- a/src/data/__tests__/auth.functions.auth-enabled.test.ts
+++ b/src/data/__tests__/auth.functions.auth-enabled.test.ts
@@ -34,6 +34,8 @@ const mockLoadAuthConfig = mock(() => ({
 mock.module('@/lib/config/auth-config', () => ({
   isAuthDisabled: () => false,
   loadAuthConfig: mockLoadAuthConfig,
+  // http redirect URI above -> plain "session" name is the active one.
+  isSecureCookie: () => false,
 }));
 
 mock.module('@/middleware/auth-middleware', () => ({
--- a/src/lib/auth/__tests__/callback-handler.test.ts
+++ b/src/lib/auth/__tests__/callback-handler.test.ts
@@ -1,11 +1,11 @@
 import { describe, it, expect, mock, afterEach } from 'bun:test';
 import {
   handleCallback,
-  buildSessionCookie,
   CLEAR_STATE_COOKIE,
   callbackGetHandler,
 } from '@/lib/auth/callback-handler';
 import type { CallbackHandlerDeps } from '@/lib/auth/callback-handler';
+import { buildSessionCookie } from '@/lib/auth/session-cookie';
 import type { OidcTokens, Role } from '@/lib/auth/types';
 import type { UserRow } from '@/lib/database/repositories/user-repository';
 
@@ -79,6 +79,7 @@ function makeDeps(overrides?: Partial<CallbackHandlerDeps>): CallbackHandlerDeps
       viewer: 'homelab-viewers',
     },
     isSecure: false,
+    sessionTtlHours: 8,
     ...overrides,
   };
 }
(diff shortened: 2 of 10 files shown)


jaredglaser:

Verified this is not obsolete: main still issues the session cookie with no Max-Age and no __Host- prefix (src/lib/auth/callback-handler.ts:38-41) and still reads the unprefixed name (src/lib/auth/resolve-user.ts:27), and issue #247 is open. The security reasoning holds: __Host- cookies cannot carry a Domain attribute, so reading only that name on https closes the subdomain-planted fixation path, and deriving isSecure from OIDC_REDIRECT_URI matches what logout-handler.ts already does, so write and read sides cannot disagree. I checked that loadAuthConfig returns process.env.OIDC_REDIRECT_URI untouched, with no trim or lowercase, so the two predicates really are evaluating the same string. The read-side conversion is also provably complete: there are exactly four non-test readers of the session cookie (src/data/auth.functions.tsx, src/lib/auth/sse-auth.ts, src/middleware/auth-middleware.ts, src/lib/auth/logout-handler.ts) and all four are converted. The tests are worth calling out too, particularly session-cookie.test.ts:105-109, which asserts that __Host-session=only-host does not satisfy a read for the plain name.

The main change I would ask for is topology: please retarget this to main. Because base.ref is claude/playwright-ci-lfyzhx and ci.yml triggers pull_request only for base main, this security fix has never run build, test, coverage or license check, and four of its eight merge conflicts (ci.yml, bun.lock, package.json, and the mock-event-source.test.ts modify/delete) are inherited from that base rather than from this change. Your own diff is 12 files and 274 lines, all under src/lib/auth, src/data/auth.functions.tsx, src/lib/config/auth-config.ts and src/middleware, with no e2e, MSW or CI file in it, so nothing here depends on the Playwright stack.

One correction to a note I would otherwise have made: isSecureCookie() is still needed after #340, because resolveUserFromCookie has no parsed auth config in scope and only imports isAuthDisabled, so the alternative is calling loadAuthConfig(), which throws when the OIDC env vars are incomplete. Two smaller items: readCookie at session-cookie.ts:44-52 is a byte-for-byte copy of parseCookie, which after rebase lives in resolve-user.ts, so please fold it rather than leave two cookie parsers in the auth path; and the migration note should say that live https sessions get one forced re-login, which is what issue #247 anticipates, rather than that there is nothing to migrate.

### Specific items to fix
- Stacked on the Playwright CI branch, so it has never run CI and inherits four unrelated conflicts
- readCookie duplicates parseCookie byte for byte; after rebase that becomes two cookie parsers in the auth path
- Single-name switch forces one re-login for every live https session, which the description does not state
- buildClearSessionCookie(false) on the AUTH_DISABLED logout path leaves a stale __Host-session cookie
- JSDoc volume in session-cookie.ts exceeds main's current CLAUDE.md rule 14

**Effort:** 2 to 3 hours: retarget to main, resolve the four #340 conflicts, fold readCookie into parseCookie, reword the migration note, trim comments.

**Obsolescence:** no. Proof: origin/main:src/lib/auth/callback-handler.ts:38-41 (no Max-Age, plain name); origin/main:src/lib/auth/resolve-user.ts:27 (parseCookie(cookieHeader, 'session')); no __Host, isSecureCookie or session-cookie.ts anywhere in main's src; issue #247 open.

- `origin/main:src/lib/auth/callback-handler.ts:38-41 (buildSessionCookie: no Max-Age, plain 'session')`
- `origin/main:src/lib/auth/resolve-user.ts:18-28 (no auth config in scope; parseCookie(header,'session') at :27)`
- `origin/main:src/lib/auth/session-manager.ts:29-30 (fixed expiry, no sliding refresh)`
- `origin/fix/f22-session-cookie-attrs:src/lib/auth/__tests__/session-cookie.test.ts:105-109 (regex-boundary regression test)`
- `origin/fix/f22-session-cookie-attrs:src/lib/auth/session-cookie.ts:44-52 (readCookie duplicates parseCookie byte for byte)`
- `git diff --stat origin/claude/playwright-ci-lfyzhx...origin/fix/f22-session-cookie-attrs -> 12 files, +274/-65, all auth`
- `git merge-tree --write-tree origin/main origin/fix/f22-session-cookie-attrs -> 8 conflicts; control run against origin/claude/playwright-ci-lfyzhx -> 4 of them are inherited`
- `mcp__github__pull_request_read get_check_runs #281 -> total_count 0; issue #247 open`

</details>

---

PR #307, opened 2026-06-21
chore: normalize Node builtin imports to the node: prefix

jaredglaser:
## What

Adds the `node:` prefix to every bare Node builtin import specifier across 26 files (`crypto`, `fs`, `fs/promises`, `path`, `stream`, `readline`, `child_process`, `url`), plus the dynamic `await import('crypto')` in `git-tokens.functions.tsx`. Specifier strings only, no logic changes (43 insertions / 43 deletions, fully symmetric).

## Why

This started from the question "should we replace `node:` APIs with Bun alternatives?" The conclusion was no: Bun implements `node:*` natively and recommends the prefixed form, and swapping `src/` to `Bun.*` APIs would couple shared code to Bun for negligible benefit. The genuinely useful cleanup was the opposite direction: most of the codebase already uses the `node:` prefix, but ~26 files used bare specifiers. This change makes them consistent.

## Background: the "run dev under Bun" idea was tried and dropped

A second change was scoped to make `bun dev` run under Bun (via `bun --bun vite`) so dev matches the prod/worker runtime. During validation it was found to break the dev server: it boots (`VITE ready`) but returns HTTP 500 on every request because Nitro's Vite dev entry wires websocket support through `crossws`, whose Node adapter rejects the Bun runtime. That isn't a quick/safe fix (disabling websockets would break real features), so the dev-runtime change was reverted and is not part of this PR. Dev stays on Node, prod stays on Bun, and `src/` remains runtime-agnostic, which is exactly why the `node:` builtins (not `Bun.*`) are the right call.

### How to reproduce the dev-under-Bun failure locally

This PR does NOT contain these changes. Apply them yourself to observe the issue:

1. In `package.json`, change the `dev` script:
   ```diff
   -    "dev": "vite dev --port 3000",
   +    "dev": "bun --bun vite dev --port 3000",
   ```
   The `--bun` flag overrides the `vite` binary's `#!/usr/bin/env node` shebang and forces the Bun runtime. Without it, `bun dev` still runs Vite under Node.

2. Make sure deps are installed (`bun run setup`), then start dev:
   ```bash
   bun dev
   ```
   It boots fine:
   ```
   $ bun --bun vite dev --port 3000
     VITE v8.0.16  ready in ~3000 ms
     ➜  Local:   http://localhost:3000/
   ```

3. Hit any route and watch it fail:
   ```bash
   curl -i http://localhost:3000/
   ```
   Returns `HTTP/1.1 500`, and the dev server stdout prints:
   ```
   error: [crossws] Using Node.js adapter in an incompatible environment.
       at nodeAdapter (node_modules/crossws/dist/_chunks/node.mjs:25:17)
       at nitro/dist/runtime/internal/vite/dev-entry.mjs:21:52
   ```

4. Revert the `package.json` change to get a working dev server again.

**Root cause:** `vite.config.ts` enables `nitro({ ..., features: { websocket: true } })`. Nitro's Vite dev entry instantiates the `crossws` Node adapter for websocket support, and that adapter throws when it detects a non-Node runtime (Bun). The web server only runs under Bun in production (`Dockerfile`: `bun .output/server/index.mjs`), where this dev-only Nitro path isn't used, so prod is unaffected.

**Possible paths to actually enable it** (out of scope here, would need investigation): a Nitro Bun/dev preset or a crossws Bun adapter, or gating `features.websocket` per-environment if dev websockets aren't needed.

## Verification

- `bun run typecheck:all` passes.
- `bun run test:all`: identical results before and after (2479 pass / 396 fail), and identical to the base commit. The pre-existing failures are unrelated environment/mock-isolation issues, not introduced here.

### DIFF
--- a/.github/scripts/triage-issues.ts
+++ b/.github/scripts/triage-issues.ts
@@ -6,8 +6,8 @@
  *   triage-summary.md: human-readable summary for the PR description
  */
 
-import { readFileSync, writeFileSync } from "fs";
-import { dirname } from "path";
+import { readFileSync, writeFileSync } from "node:fs";
+import { dirname } from "node:path";
 
 const MAX_ISSUES = Math.min(Math.max(1, Number(process.env.MAX_ISSUES ?? "10") || 10), 50);
 const INPUT_FILE = process.env.INPUT_FILE ?? "sonar-issues.json";
--- a/agent/src/__tests__/stacks.test.ts
+++ b/agent/src/__tests__/stacks.test.ts
@@ -8,8 +8,8 @@ import {
   handleStackStatus,
   parseContainerNames,
 } from '../routes/stacks';
-import { mkdirSync, rmSync, existsSync, readFileSync, statSync, writeFileSync } from 'fs';
-import { join } from 'path';
+import { mkdirSync, rmSync, existsSync, readFileSync, statSync, writeFileSync } from 'node:fs';
+import { join } from 'node:path';
 
 beforeAll(() => {
   console.error = mock(() => {});
--- a/scripts/download-icons.ts
+++ b/scripts/download-icons.ts
@@ -6,9 +6,9 @@
  *
  * @see https://github.com/homarr-labs/dashboard-icons
  */
-import { mkdir, writeFile, readdir, cp, rm } from 'fs/promises';
-import { existsSync } from 'fs';
-import { execSync } from 'child_process';
+import { mkdir, writeFile, readdir, cp, rm } from 'node:fs/promises';
+import { existsSync } from 'node:fs';
+import { execSync } from 'node:child_process';
 
 const ICONS_DIR = './public/icons';
 const REPO_URL = 'https://github.com/homarr-labs/dashboard-icons.git';
--- a/src/data/git-tokens.functions.tsx
+++ b/src/data/git-tokens.functions.tsx
@@ -10,7 +10,7 @@ export const createGitToken = createServerFn()
   .handler(async ({ data, context }): Promise<{ token: string }> => {
     requireRole('admin')(context.user);
 
-    const { randomBytes } = await import('crypto');
+    const { randomBytes } = await import('node:crypto');
     const { databaseConnectionManager } = await import('@/lib/clients/database-client');
     const { loadDatabaseConfig } = await import('@/lib/config/database-config');
     const { loadMasterKeyring } = await import('@/lib/crypto/master-key');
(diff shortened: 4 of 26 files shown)


jaredglaser:
## Triage verdict: Merge (confidence: high)

Verified the diff is purely specifier strings: every added and removed line is an import statement differing only by the node: prefix, including both quote styles and the one dynamic await import('crypto') in src/data/git-tokens.functions.tsx. No logic changes, and I checked the one way a rename could bite, which is a bundler external or exclude keyed on the old string; vite.config.ts:57 and :77 externals and :60 optimizeDeps.exclude name only npm packages (dockerode, ssh2, docker-modem, ssh2-streams, undici, cpu-features, @tanstack/start-server-core), never a builtin.

This is not obsolete either: main still has about 41 bare builtin specifiers, for example src/lib/database/migrate.ts:3-5, src/lib/auth/login-handler.ts:1 and src/lib/parsers/text-parser.ts:1-2. I materialized the merge tree and re-counted: exactly 4 remain afterwards, in stack-repo-layout.test.ts:2-3 and stack-service.test.ts:18-19, and both are regressions main introduced after this branch was cut (stack-repo-layout.test.ts arrived with ebb8f4f on 2026-07-22), so please fix those 4 lines during the rebase or open a follow-up. It merges with zero conflicts despite being 45 commits behind.

Two notes: the '2479 pass / 396 fail' figure in the description is an environment artifact rather than a real baseline, since .bun-version pins 1.3.14 while the authoring environment ran 1.3.11 and #361 verified main at 0 fail on the pinned version, so please strike it from the body; and since the repo has no linter, a CI grep step is the only thing that will stop this drifting back, which it demonstrably did four times in five weeks.

### Specific items to fix
- 4 bare specifiers regressed onto main after this branch was cut and are not covered by the diff
- No linter or CI guard, so the drift will recur; it already did four times in five weeks
- PR body quotes a test baseline that is an environment artifact of an unpinned Bun version

**Effort:** under an hour: fix the 4 regressed lines, strike the test-count line from the body, merge.

**Obsolescence:** no. main still carries roughly 41 bare builtin import specifiers.

- `git diff origin/main...origin/claude/node-vs-bun-modules-f8qgep -> 43 insertions / 43 deletions, all import specifier lines`
- `origin/main:vite.config.ts:57,77 (externals) and :60 (optimizeDeps.exclude) name only npm packages, no builtins`
- `origin/main:src/lib/database/migrate.ts:3-5, src/lib/auth/login-handler.ts:1, src/lib/parsers/text-parser.ts:1-2 (bare specifiers still present)`
- `merge tree materialized -> 4 bare specifiers remain: stack-repo-layout.test.ts:2-3, stack-service.test.ts:18-19`
- `commit ebb8f4f (2026-07-22) introduced stack-repo-layout.test.ts after this branch was cut`
- `git merge-tree --write-tree origin/main origin/claude/node-vs-bun-modules-f8qgep -> 0 conflicts`
- `.bun-version pins 1.3.14; commit 8fbae76 (#361) verified main at 0 test failures`

</details>

---

PR #295, opened 2026-06-15
ci: gate the release jobs on e2e and migration verification

jaredglaser:
Originally "add the e2e job", but that job landed with #293. What is left here is the release gating, kept separate so the policy change is reviewed on its own.

## What it does

Two lines in `.github/workflows/ci.yml`:

- `publish` (Docker images) gains `e2e` and `migrations` in `needs:`.
- `deploy-demo` (GitHub Pages) gains `e2e`.

`deploy-demo` is deliberately **not** gated on `migrations`: it is a static build with no database, so a migration failure says nothing about whether the demo is publishable.

Without this, both release jobs depended only on `build-test`, `agent` and `agent-updater`, so a red e2e run or a broken migration could not stop a publish.

## Why it is worth the latency

Accepted trade-off: releases now wait on the slowest of the two added jobs, and a flaky e2e run can block a publish until it is re-run. The e2e job is bounded at `timeout-minutes: 20` so it cannot hang a release indefinitely.

## Verification

Both job ids referenced in `needs:` exist in the workflow on `main`: `e2e` at `ci.yml:75` and `migrations` at `ci.yml:246`. CI is green on this head, with `publish` and `deploy-demo` correctly reporting as skipped (they run only on `push`, not on pull requests, so the new edges are exercised on merge to `main`).

## History

This branch was originally stacked on #293's branch. #293 was squash-merged, which orphaned it, so it was rebased onto `main` with `restack.sh`; only its own two commits remain.

Issue #308, opened 2026-06-21
e2e: real-server Playwright lane for the session cookie contract (Max-Age, __Host-, single-name read)


jaredglaser:
## Context

#281 hardens the auth session cookie:
- `Max-Age` from `SESSION_TTL_HOURS`, so the cookie survives browser restarts for as long as the DB session is valid.
- `__Host-session` + `Secure` on https; plain `session` on http (local dev).
- Single-name reads keyed on `isSecure` (derived from `OIDC_REDIRECT_URI` via `isSecureCookie()`): on https the plain name is ignored, so a `session` cookie planted by a subdomain cannot fixate a session past the `__Host-` guard.

Unit tests run in Happy-DOM, which reproduces the browser rule that `Set-Cookie` is a forbidden response header: `response.headers.get('Set-Cookie')` returns `null`. So #281 asserts attribute strings against the pure builder functions and asserts only status/`Location` on the handler `Response`. That covers formatting and parse logic, but not the part only a real browser can prove: that the browser accepts `__Host-`, stores the cookie with the right lifetime, sends it back, and that logout clears it.

## Why the current Playwright harness cannot cover this

The harness from #293 serves two static production builds over MSW (`playwright.config.ts`). There is no real backend: `/_serverFn/*` and `/api/*` are answered in-browser by the MSW service worker, and auth is faked with a `getSession` override (test-plan flow 17). Two hard blockers:

1. No real `/api/auth/callback` runs, so no real `Set-Cookie` is ever emitted.
2. MSW is in-browser JS and cannot set `HttpOnly` cookies; the session cookie is `HttpOnly`.

This needs a separate lane with a real server, not the MSW harness.

## What genuinely needs a browser vs. what does not

- Attribute-string correctness (name, `Max-Age`, `Secure`, `HttpOnly`, `SameSite`, `Path`) does not need a browser. The pure-string unit tests in #281 already cover it, and a Bun integration test against a real `Bun.serve` could assert `response.headers.getSetCookie()` directly if Response-level coverage is wanted.
- Browser acceptance and round-trip does need a browser. `__Host-` is silently rejected if malformed, so only a real cookie jar proves it was accepted, persisted for `Max-Age`, sent on the next request, and cleared on logout. That is the unique value of a Playwright lane.

### Phase 1 (no browser, keep / optionally add)
[...]

PR #310, opened 2026-06-21
test(e2e): real-server auth lane for the session cookie contract

jaredglaser:
Re-cut onto `main` on 2026-07-27. Previously stacked behind #293, #295 and #281; all three have merged, so this is now a single commit against `main` with no stack.

Adds the end-to-end lane that proves the OIDC session cookie contract in a real browser.

## Why a separate lane

The MSW harness (#293) cannot test the session cookie. It serves static builds with an in-browser mock backend, a service worker cannot emit an `HttpOnly` `Set-Cookie`, and auth there is faked through a `getSession` override. The cookie attributes are equally out of reach for Happy-DOM unit tests, since `Set-Cookie` is a forbidden response header. They need a real server and a real browser.

## What it does

- `playwright.auth.config.ts` defines two projects: `auth-http` (3201) and `auth-https` (3202, self-signed TLS with `ignoreHTTPSErrors`). The MSW `app` project ignores the auth specs, so the lanes never cross.
- Boots the real, non-MSW Nitro build against a real **Pocket ID** and a throwaway **Timescale** database. The cookie name is driven by `OIDC_REDIRECT_URI`, so the two servers exercise the `session` and `__Host-session` modes.
- Login uses Pocket ID's one-time access token, minted per test through the admin API, so there is no passkey or WebAuthn automation.
- https is reached through a small Bun TLS proxy in front of an internal Nitro instance, because `https://localhost` is a secure context in Chromium and is therefore required for `__Host-` and `Secure` to be accepted at all.

## Assertions

- **http**: the real callback `Set-Cookie` carries `Max-Age`, `HttpOnly`, `SameSite=Lax`, `Path=/` and no `Secure`; the browser stores it with a real expiry; the session survives into a fresh context; logout clears it.
- **https**: `__Host-session` with `Secure` is accepted and stored, which is itself the proof, since a malformed `__Host-` cookie would be silently dropped. A planted plain `session` cookie does **not** authenticate, covering the single-name read. Logout clears it.

## CI

New `e2e-auth` job with `timescale/timescaledb` and `ghcr.io/pocket-id/pocket-id:v2` service containers, running `bun run test:e2e:auth`.

`publish` now requires `e2e-auth` alongside the `e2e` and `migrations` gates added by #295. `deploy-demo` deliberately does **not**: it is a static build with no database and no OIDC, so a Pocket ID container failing says nothing about whether the demo is publishable. That is the same reasoning that kept `migrations` off `deploy-demo`.

The job is bounded at `timeout-minutes: 20`, matching `e2e`. It gates `publish`, and an unbounded job that boots two service containers can block a release indefinitely.

## What the re-cut changed

The branch carried three commits that have since merged in evolved form, so only the auth-lane commit was replayed. The cookie contract these tests assert (`__Host-session`, `Max-Age`, `buildClearSessionCookie`) was confirmed present on `main` before the superseded commit was dropped.

Specs were renamed from `*.authhttp.spec.ts` to `*.authhttp.e2e.ts`, with the matchers and ignores updated. This branch predates the rename #293 made: `bun test` collects `*.spec.ts` and dies on Playwright's `test()` with no failing assertion, so keeping the old suffix would have reintroduced that. `docs/playwright-test-plan.md` was deleted on `main`; the lane's description belongs in the flow inventory (#384) rather than resurrecting the file.

## Verification

Docker is not available in the environment this was re-cut in, so the full Pocket ID and Postgres flow is verified by the `e2e-auth` job in CI. Verified locally: `bun run typecheck` clean; `bun test` collects no e2e specs; the MSW lane still lists 21 tests in 9 files with the auth specs correctly excluded; the auth lane lists 6 tests across both projects.

## Summary by CodeRabbit

* **New Features**
  * Added end-to-end authentication coverage for HTTP and HTTPS deployments.
  * Verified secure session cookie behavior, persistence, logout, and rejection of forged cookies.
  * Added automated setup for authentication services and test databases.

* **Tests**
  * Added dedicated Playwright authentication test projects and CI execution.
  * Authentication test reports are now uploaded as build artifacts.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,6 +114,89 @@ jobs:
           path: playwright-report/
           retention-days: 14
 
+  e2e-auth:
+    name: E2E Auth (Playwright + Pocket ID)
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    permissions:
+      contents: read
+
+    # Real-server lane: runs the actual Nitro build against a real Pocket ID and
+    # Postgres so the session cookie (Max-Age, __Host-/Secure, single-name read)
+    # is verified in a real browser. The MSW lane above cannot: it mocks the
+    # backend and cannot emit HttpOnly Set-Cookie.
+    services:
+      postgres:
+        image: timescale/timescaledb:latest-pg16
+        env:
+          POSTGRES_DB: homelab_test
+          POSTGRES_USER: homelab
+          POSTGRES_PASSWORD: homelab
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd "pg_isready -U homelab"
+          --health-interval 5s
+          --health-timeout 5s
+          --health-retries 10
+
+      pocket-id:
+        image: ghcr.io/pocket-id/pocket-id:v2
+        env:
+          APP_URL: http://localhost:1411
+          # Throwaway values for an ephemeral CI instance; not real secrets.
+          ENCRYPTION_KEY: e2e-pocket-id-encryption-key-not-a-secret
+          STATIC_API_KEY: e2e-pocket-id-static-api-key-not-a-secret
+          DISABLE_RATE_LIMITING: "true"
+          ANALYTICS_DISABLED: "true"
+        ports:
+          - 1411:1411
+        options: >-
+          --health-cmd "wget --spider -q http://localhost:1411/healthz || exit 1"
+          --health-interval 5s
+          --health-timeout 5s
+          --health-retries 20
+          --health-start-period 20s
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+
+      - name: Setup Bun
+        uses: oven-sh/setup-bun@v2
+
+      - name: Install dependencies
+        run: bun install --frozen-lockfile
+
+      - name: Install Playwright browser
+        run: bunx playwright install --with-deps chromium
+
+      - name: Generate a throwaway master key
+        run: echo "MASTER_KEY=$(openssl rand -base64 32)" >> "$GITHUB_ENV"
+
+      - name: Run auth e2e tests
+        run: bun run test:e2e:auth
+        env:
+          CI: '1'
+          POSTGRES_HOST: localhost
+          POSTGRES_PORT: '5432'
+          POSTGRES_DB: homelab_test
+          POSTGRES_USER: homelab
+          POSTGRES_PASSWORD: homelab
+          POSTGRES_SSL: 'false'
+          POCKET_ID_URL: http://localhost:1411
+          POCKET_ID_API_KEY: e2e-pocket-id-static-api-key-not-a-secret
+
+      - name: Upload Playwright report
+        if: ${{ !cancelled() }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: playwright-report-auth
+          path: playwright-report/
+          retention-days: 14
+
   agent:
     name: Agent Test & License Check
     runs-on: ubuntu-latest
@@ -332,7 +415,7 @@ jobs:
   publish:
     name: Publish Docker Images
     runs-on: ubuntu-latest
-    needs: [build-test, agent, agent-updater, e2e, migrations]
+    needs: [build-test, agent, agent-updater, e2e, e2e-auth, migrations]
     if: github.event_name == 'push'
     permissions:
       contents: read
--- a/e2e/auth.authhttp.e2e.ts
+++ b/e2e/auth.authhttp.e2e.ts
@@ -1,0 +1,57 @@
+import { test, expect } from '@playwright/test';
+import { login, setCookieValues } from './auth/login';
+
+/**
+ * http deployment: the app issues the plain `session` cookie (no __Host-, no
+ * Secure). Asserts the real Set-Cookie header and the browser's stored cookie,
+ * then proves persistence (the cookie is sent back and validated) and that
+ * logout clears it. baseURL is http://localhost:3201 (auth-http project).
+ */
+const BASE = 'http://localhost:3201';
+
+test('login issues a plain session cookie with Max-Age and HttpOnly, no Secure', async ({
+  page,
+  context,
+}) => {
+  const callback = await login(page, BASE);
+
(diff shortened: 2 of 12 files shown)


jaredglaser:
## Triage verdict: Needs your decision (confidence: medium)

The framing in #308 is right and the engineering here is the most careful in this group: Happy-DOM makes Set-Cookie unreadable and MSW cannot emit HttpOnly, so __Host- acceptance really does need a browser with a real cookie jar.

What I would put back to you is that #308 already specifies a cheaper design than this PR builds. Its Plan step 2 describes a stub OIDC provider (a small Bun.serve satisfying discovery, /authorize redirecting straight back with a code and the round-tripped state, /token returning a signed id_token, and /jwks), and step 1 says to add a third project to the existing playwright.config.ts. This PR instead boots ghcr.io/pocket-id/pocket-id:v2 and timescale/timescaledb:latest-pg16 as service containers with an admin-API seed, adds a second Playwright config, and makes both publish and deploy-demo depend on the result, so a flake anywhere in that chain blocks GHCR publication on a push to main, and the floating timescaledb tag means an upstream image push can break main with no repo change. Choosing between the two is your call; the plan you already wrote is the cheaper one.

Regardless of which way that goes, several concrete fixes. The forged-cookie test at auth.authhttps.spec.ts:33-42 asserts only that the URL becomes /login, and src/hooks/useAuth.ts navigates to /login when getSession returns null, when it throws, and when the network fails, so a down database or a broken build passes this test; it needs a positive control that plants the correct __Host-session value and asserts it does authenticate. Both logout tests look broken as written: they read Set-Cookie off page.goto('/api/auth/logout'), but handleLogout returns 302 with Location set to the IdP end_session_endpoint and Playwright resolves goto with the response of the last redirect, so with a live Pocket ID the cookie is not on the response you read and expect(undefined).toContain('Max-Age=0') throws; use the waitForResponse pattern login() already uses. login() itself at e2e/auth/login.ts:68 waits for exactly baseURL + '/', but src/routes/index.tsx throws redirect({ to: '/docker' }) in beforeLoad so '/' is never a resting state, and that helper is used by five of the six tests. scripts/run-migrations.ts is never invoked by anything, since test:e2e:auth runs the seed script which does migrations inline. The lane also runs the app under node (scripts/e2e-auth-server.ts:51 spawns node) while production runs bun (Dockerfile:29), which is a poor match for a lane whose entire subject is Set-Cookie emission. And scripts/e2e-auth-seed.ts's docblock says oidc.json carries oneTimeLoginUrl while the code writes userId, which e2e/auth/login.ts depends on.

On sequencing: the only commit unique to #281 touches no e2e, MSW or CI file, so retargeting #281 straight at main and re-pointing this PR at it would cut the trial-merge conflicts here from 8 paths to a much smaller set. Note also that no CI has ever run on this PR.

### Specific items to fix
- Builds a more expensive lane than issue #308's own plan, which specifies a stub OIDC provider and a third project in the existing config
- Both logout tests read Set-Cookie off a goto whose last redirect is the IdP, so cleared is undefined and the assertion throws
- login() waits for baseURL + '/', which src/routes/index.tsx never rests on; used by five of six tests
- The forged-cookie test passes on any failure (useAuth routes to /login on null, throw, or network error); no positive control
- publish and deploy-demo gated on a job booting two service containers and an external IdP, with a floating timescaledb tag
- Runs the app under node while production runs bun, in a lane whose subject is header emission
- scripts/run-migrations.ts is dead code; seed docblock contradicts what the script writes

**Effort:** if the current design is accepted, 1 to 2 days to fix the logout and login helpers, add a positive control, ungate publish, remove the dead script, and get one green run. If #308's stub-IdP design is preferred, closer to a rewrite of the lane infrastructure, roughly 2 to 3 days, but it drops the service containers and the second config.

**Obsolescence:** no, and it cannot become obsolete until #281 merges. Proof: no __Host-session anywhere in main's src; origin/main:src/lib/auth/session-cookie.ts does not exist; origin/main:src/lib/auth/logout-handler.ts still writes a plain session= cookie; issue #308 open.

- `mcp__github__issue_read #308 -> open; Plan step 1 specifies a third project in playwright.config.ts, step 2 specifies a stub Bun.serve OIDC provider`
- `origin/main has no __Host-session reference in src; origin/main:src/lib/auth/logout-handler.ts writes a plain session= cookie`
- `origin/claude/e2e-auth-cookie-lane:src/lib/auth/logout-handler.ts:60-68 (302 with Location oidcLogoutUrl and the clearing Set-Cookie on that 302) vs e2e/auth.authhttp.spec.ts:52-54 and auth.authhttps.spec.ts:47-49`
- `e2e/auth/login.ts:68 (waitForURL(baseURL + '/')) vs origin/main:src/routes/index.tsx:3-7 (beforeLoad throws redirect to /docker)`
- `origin/main:src/hooks/useAuth.ts (navigates to /login on null, on throw, and on network failure), refuting 'not fakeable'`
- `git merge-tree --write-tree origin/main origin/claude/e2e-auth-cookie-lane -> 8 CONFLICT lines; 4 of them from #281 colliding with #340`
- `scripts/run-migrations.ts is referenced nowhere outside itself on this branch`
- `origin/main:Dockerfile:29 CMD ['bun', '.output/server/index.mjs'] vs scripts/e2e-auth-server.ts:51 Bun.spawn(['node', serverEntry])`
- `get_check_runs #310 -> total_count 0`

</details>

---

.github/workflows/ci.yml
```diff
@@ -127,7 +127,10 @@ jobs:
     # backend and cannot emit HttpOnly Set-Cookie.
     services:
       postgres:
-        image: timescale/timescaledb:latest-pg16
+        # Pinned by digest, matching the migrations job's convention below: a
+        # floating tag (latest-pg16) lets an upstream image push break main with
+        # no commit in this repo. Bump tag + digest together.
```
jaredglaser: this verbose comment is pointless

.github/workflows/ci.yml
```diff
@@ -141,7 +144,8 @@ jobs:
           --health-retries 10
 
       pocket-id:
-        image: ghcr.io/pocket-id/pocket-id:v2
+        # Pinned by digest (linux/amd64 manifest of the v2 tag at adoption time).
```
jaredglaser: again here

PR #354, opened 2026-07-23
feat(stacks): toast real deploy outcomes, not just request acceptance

jaredglaser:
## Why

The deploy server function awaits the entire pipeline, including the agent's `docker compose up`, and receives the final status and logs, then returns only `{ deployId }`. A deploy that dispatched and failed still resolved the mutation and toasted "deploy triggered successfully". Async paths (git push deploys, watchdog timeouts, delete-teardown, other tabs) produced no outcome signal at all; the user had to open deploy history to learn what happened.

## What

Two layers, deduplicated by a deployId-keyed gate:

**Real status from server functions.** `triggerDeploy`/`resumeDeploy` return `{ deployId, status, logs }`. `resumePendingDeploy` no longer throws on a failed dispatch; it returns the status for the caller to toast. The UI toasts the actual outcome ("Deploy of plex succeeded" / "Deploy of plex failed: <first line of logs>"), which works even with SSE disconnected.

**Enriched `deploy_change` NOTIFY.** The payload gains optional `deployId`/`status`/`action`/`trigger`/`message` (sanitized: first line, control chars stripped, surrogate-safe 200-char truncation), threaded through the broadcast service, the route's serialize whitelist, and the zod wire schema (all optional, so version skew is safe). `useStackStatus` toasts terminal outcomes while any stacks route is mounted. This covers git push deploys, watchdog-failed deploys, startup recovery, delete-teardown after navigation, and deploys from other sessions. Outcomes ride only on live NOTIFY forwards, never init snapshots, so page loads cannot replay old toasts.

**Dedup and races.** A singleton atomic test-and-set gate keyed by deployId ensures one toast regardless of which source lands first. Approve/reject claim the gate in `onMutate` (their deployId is the mutation argument), so their local toast deterministically beats the SSE echo; without that, rejecting a deploy could show an error-styled "Deploy failed: Manually rejected" toast. Both paths share one formatter with identical truncation so copy never differs by race winner. Suppressed: `pending`, `in_progress`, and `no_change` from non-UI triggers.

Delete-with-teardown now shows a neutral "Teardown started" info toast (the client navigates away; the outcome arrives via SSE on the stacks list). Also fixed the stale doc comment claiming `deleteStack` "returns immediately" (it awaits the teardown server-side).

## Tests

New formatter matrix + gate semantics suite; updates across deploy-repository (payload shape, sanitization, RETURNING columns), pipeline (outcome per branch), watchdog, startup recovery, broadcast service, SSE schema, stack-service/mappers, functions, useStackStatus (toast-once, duplicate-suppressed, legacy-payload-no-toast, init-never-toasts), StackEditorForm, and DeployHistoryList/Row.

https://claude.ai/code/session_015VS9bkwReJgTDAP5Z9eFjq

---
_Generated by [Claude Code](https://claude.ai/code/session_015VS9bkwReJgTDAP5Z9eFjq)_

## Summary by CodeRabbit

* **New Features**
  * Deployment, approval, and rollback notifications now show outcome-specific success or failure messages.
  * Failure notifications include helpful, sanitized details when available.
  * Live deployment updates now include structured outcome information and avoid duplicate notifications.

* **Bug Fixes**
  * Improved reporting for failed deployments, approvals, rollbacks, recoveries, and timed-out operations.
  * Removed duplicate teardown notifications when completion is reported through live updates.
  * Pending deployment resumes now surface their actual result instead of hiding failures.

### DIFF
--- a/src/hooks/useStackStatus.ts
+++ b/src/hooks/useStackStatus.ts
@@ -35,20 +35,8 @@ export function useStackStatus() {
   const handleData = useCallback((data: StackSSEMessage) => {
     if (isDeployChanged(data)) {
       setDeployVersion((v) => v + 1);
-      if (
-        data.deployId !== undefined &&
-        data.status !== undefined &&
-        data.action !== undefined &&
-        data.trigger !== undefined &&
-        deployToastGate.shouldToast(data.deployId)
-      ) {
-        const outcome = formatDeployOutcome({
-          stack: data.stack,
-          action: data.action,
-          status: data.status,
-          trigger: data.trigger,
-          message: data.message,
-        });
+      if (data.outcome !== undefined && deployToastGate.shouldToast(data.outcome.deployId)) {
+        const outcome = formatDeployOutcome({ stack: data.stack, ...data.outcome });
         if (outcome) showToast(outcome.message, outcome.severity);
       }
       return;
--- a/src/hooks/useStackStatus.ts
+++ b/src/hooks/useStackStatus.ts
@@ -1,9 +1,13 @@
 import { useCallback, useState } from 'react';
 import { useSseChannel } from '@/hooks/useSseChannel';
 import { stackStatusChannel, type StackSSEMessage } from '@/lib/sse/channels/stack-status';
+import { useToast } from '@/hooks/toastAtom';
+import { deployToastGate, formatDeployOutcome } from '@/lib/stacks/deploy-outcome-toast';
 import type { StackStatusEntry } from '@/types/stacks';
 
-function isDeployChanged(data: StackSSEMessage): data is { type: 'deploy_changed'; stack: string; host: string } {
+type DeployChangedMessage = Extract<StackSSEMessage, { type: 'deploy_changed' }>;
+
+function isDeployChanged(data: StackSSEMessage): data is DeployChangedMessage {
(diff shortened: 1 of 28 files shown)


src/hooks/useStackStatus.ts
```diff
@@ -26,10 +30,27 @@ function shallowEqualContainers(
...
   const handleData = useCallback((data: StackSSEMessage) => {
     if (isDeployChanged(data)) {
       setDeployVersion((v) => v + 1);
+      if (
+        data.deployId !== undefined &&
+        data.status !== undefined &&
+        data.action !== undefined &&
+        data.trigger !== undefined &&
+        deployToastGate.shouldToast(data.deployId)
+      ) {
```
jaredglaser: this looks brittle

jaredglaser: Agreed. The brittleness was that the wire schema modeled the five outcome fields as independently optional, so the all-or-nothing invariant lived only in this four-way `!== undefined` conjunction. Fixed in 11779d7: the fields are now nested under a single optional `outcome` object whose inner fields are required, so zod rejects a partial outcome at the boundary (new test: "rejects an outcome missing a required field (deployId)") and this check collapses to `data.outcome !== undefined && gate.shouldToast(data.outcome.deployId)`.

---

PR #357, opened 2026-07-23
feat(stacks): show container ports and mounts in the Containers tab

jaredglaser:
## Why

The Docker page's detail panel gained ports/mounts in #355; the stack editor's Containers tab still showed only name/status/image. This surfaces the same data per container row.

Stacked on #355 (`claude/web-ports-mounts`).

## What

- `StackContainer` carries `ports`/`mounts`; the stack-status wire schema gains both with `.default([])` (skew-safe). The `deploy_changed` variant of that schema is deliberately untouched (PR #354 modifies it).
- The stack-status broadcast service's NOTIFY handler now merges instead of overwriting its in-memory container entry: the NOTIFY payload carries ports but never mounts (migration 026's size policy), so a naive overwrite would blank every container's mounts on its first event. Mounts always come from the previous entry; ports come from the payload with a fall-back to the previous entry when the field is absent (pre-026 skew).
- `StackContainersPanel` rows render port chips inline after the image (reusing #355's `dedupeWildcardPorts`/`formatPortMapping` helpers, unmodified) and mounts behind a drive icon with a tooltip listing `source -> destination [ro]` per mount. Rows stay single-line; nothing renders when both arrays are empty.

## Tests

Broadcast service: init snapshot carries the fields; upsert-with-ports preserves previous mounts; upsert-without-ports preserves both; new container gets payload ports and empty mounts. Schema round-trips with and without the fields. Panel: chip rendering/dimming, tooltip content, hidden empty state. All touched files green individually; `typecheck:all` clean.

https://claude.ai/code/session_015VS9bkwReJgTDAP5Z9eFjq

---
_Generated by [Claude Code](https://claude.ai/code/session_015VS9bkwReJgTDAP5Z9eFjq)_

## Summary by CodeRabbit

* **New Features**
  * Stack container views now show published and unpublished ports inline.
  * Added overflow handling for containers with many ports.
  * Added mount indicators with tooltips showing mount locations, volume names, and read-only status.
  * Container status updates now retain port and mount information reliably.

* **Bug Fixes**
  * Improved compatibility when status updates omit port or mount details.
  * Hidden container controls when no associated service is available.


src/lib/stacks/stack-status-broadcast-service.ts
```diff
@@ -363,7 +358,16 @@ export class StackStatusBroadcastService {
           byContainer = new Map();
           this.stackContainers.set(sk, byContainer);
         }
-        byContainer.set(`${inv.host}/${inv.containerId}`, inv);
+        const containerKey = `${inv.host}/${inv.containerId}`;
+        const prev = byContainer.get(containerKey);
+        // NOTIFY (migration 026) never carries mounts, so the previous entry's mounts always win;
+        // ports only fall back to the previous entry when the field is absent (pre-026 skew) or
+        // null (oversized-payload size guard), not when it's genuinely [].
```
jaredglaser: this comment seems pointless

jaredglaser: Trimmed the three-line block to a single line in c6a06fc. Kept the one non-obvious bit (NOTIFY omits mounts and nulls oversized ports per migration 026, so the previous entry's values win, but a genuine `[]` is honored) since the `parsed.ports != null` guard was a prior bug locus. Say the word and I'll drop it entirely.

---

PR #370, opened 2026-07-25
feat(db): three-tier stats retention cascade with hourly kept forever

jaredglaser:
Stacked on #276 (target that branch, not `main`).

## Summary

Implements the retention model the Data Retention settings UI has always described but nothing ever applied: **raw to 1-minute to 1-hour, with the hourly tier kept forever**, across all three stats hypertables.

The original #276 hardcoded a conflicting two-tier scheme (30 days raw, a 1-minute aggregate kept 365 days, no hourly tier) that ignored the three sliders on the settings page and guaranteed data loss at one year. This replaces it.

| Tier | Chunk | Compress after | Retention |
|---|---|---|---|
| raw 1 Hz | 1 hour | 6 hours | 24 hours |
| `_1m` | 1 day | 7 days | 30 days |
| `_1h` | 7 days | 30 days | never |

Applied to `docker_stats` (030), `zfs_stats` (031) and `proxmox_stats` (032). Shipping raw retention for zfs and proxmox with no rollup, as #276 did, was the single irreversible thing in it.

## Three invariants

**Hourly is built on the minute aggregate, not on raw.** If `_1h` read raw, raw's 24-hour retention would become its recovery window: a refresh job down for two days could never materialize the gap, and a refresh recomputing a region whose raw chunks are gone would delete the aggregate for that region. Sourcing from the 30-day `_1m` gives 30 days of slack.

**Each refresh window sits strictly inside its source's retention.** `_1m` over raw uses `start_offset 6 hours`; `_1h` over `_1m` uses `start_offset 3 days`.

**Weighted averages, not average-of-averages.** `AVG()` over the minute `AVG()`s is only correct when every minute has identical sample counts. `_1m` carries `count(*) AS sample_count`; `_1h` materializes `SUM(metric * sample_count)` and `SUM(sample_count)`; a plain view divides, since a cagg definition cannot divide two aggregates. On a realistic sparse hour this is the difference between `10.27027027027027` and `11.33333`.

## Per-source column treatment

Group keys mirror each table's existing `compress_segmentby`.

- **docker**: `last()` on `container_name`/`image`; `AVG()` on all eight metrics, which are gauges or pre-computed rates.
- **zfs**: `last()` on `entity_type`/`indent`; `AVG()` on the seven metrics.
- **proxmox**: `AVG()` on exactly four true gauges (`cpu`, `mem`, `disk`, `storage_avail`). **`last()` on `netin`, `netout` and `uptime`**: these arrive from the Proxmox API as cumulative totals with no delta applied by the collector, and `GuestSection.tsx` renders them as raw byte totals. Averaging a monotonic counter yields the bucket midpoint, which is meaningless and looks plausible. Also `last()` on the capacity constants `max_cpu`/`max_mem`/`max_disk` and on `status`, `node`, `entity_name`, `vmid`, `storage_type`, `storage_content`, `storage_shared`, `cluster_version`. A test asserts `AVG(netin)` never appears and that exactly four metrics are weighted.

## Backfill and retention run as a startup task, not in the migrations

The migrations now contain only transactional DDL: the chunk interval changes, the two continuous aggregates (`WITH NO DATA`), the `_1h_avg` dividing view, the refresh policies and the compression settings and policies. Every migration file runs inside one transaction again.

The initial backfill and the retention policies moved to `runStatsRollupBackfill` (`src/lib/database/stats-rollup-backfill.ts`), fired from `initServer()` next to `startDeployRecovery()`. Per source, in order:

1. `SELECT min(time) FROM <table>` gives the real backfill range. An empty table skips straight to step 3.
2. Refresh `_1m` in bucket-aligned 2-day windows from that minimum up to `now()`, then `_1h` in 6-day windows. `_1m` completes first, since `_1h` reads it. Progress is logged per window at `console.info`; a large backlog can take a while.
3. **Only after both tiers succeed**, add that source's retention policies (`_1m` 30 days, raw 24 hours, `if_not_exists => TRUE`). `_1h` gets none.

If any window fails, the failure is logged at `console.error`, retention is not added for that source, and the next source proceeds. The next boot retries against history that is still there. This is strictly safer than doing it in SQL, where statement ordering inside a file was the only thing standing between a failed backfill and a retention policy deleting rows that were never rolled up.

The whole task takes its own `pg_try_advisory_lock` (key `0x686c6d32`, session-scoped, on one dedicated `PoolClient`, released in a `finally`). A second web instance, or the worker booting alongside, skips rather than blocking. If a source's aggregates do not exist yet (migrations still pending in the worker on a first upgrade boot), the source is skipped with an info log and picked up on a later boot; the refresh policies created by the migration keep new data rolled up in the meantime, and no retention is in force, so nothing is lost.

`server-init.ts` is web-only. A worker-only deployment is not a supported configuration (the web container is the product), but if someone ran one, the aggregates and their refresh policies would still be created by the migrations and ongoing data would roll up; only the historical backfill and retention would never run, which fails toward keeping data rather than dropping it.

### Why this replaces the previous approach

`CALL refresh_continuous_aggregate` cannot run inside a transaction block, and node-postgres sends a multi-statement string over the simple query protocol, which wraps the batch in an implicit transaction. Doing it in SQL therefore required a `-- migrate:no-transaction` mode and roughly 70 lines of hand-rolled SQL lexer (`splitSqlStatements`) in `migrate.ts`. That lexer already had two gaps (nested block comments, `E''` backslash escapes) that today's migrations happen not to reach. All of it is gone: `migrate.ts` and its test are byte-identical to their state on the base branch, and each `CALL` is now its own `client.query`, so there is nothing to split by construction.

Two previously listed concerns drop out as a result:

- The non-idempotent `ALTER MATERIALIZED VIEW ... SET (timescaledb.compress...)` no longer matters: the file is transactional again, so a mid-file failure rolls back and the migration re-runs from a clean state.
- **#369's forced-replay check becomes moot for this PR**: no migration carries the no-transaction directive any more, so there is no file that re-runs partially applied.

## Routing takes the coarser of span and age

`resolveStatsTier(table, windowSeconds, windowStartAgeSeconds = windowSeconds)` computes a tier from each dimension and takes the maximum. The default parameter leaves the three `NOW()`-anchored `*History(seconds)` methods unchanged, where span equals age.

`getDockerStatsForContainer` passes real age. Without this, a short window pointed at a distant past (a panned or zoomed detail chart) routed to raw by span, and raw is gone after 24 hours, so the chart rendered empty. Since the dashboard chart is capped at 1800s and cannot reach the aggregates, this view is their only consumer, so the bug would have made the tiers unreachable in the one place they exist to serve.

`getContainerInfo` now reads from `docker_stats_1m` rather than raw, so a container idle more than 24 hours keeps its name and image.

## Settings

The three Data Retention controls are set to the hardcoded values and disabled; the hourly control becomes a static "Kept forever" row, since forever is not representable on a 1-365 slider. Help text updated. The `retention` settings plumbing is left in place for a future PR that makes these operator-configurable.

## Known limitations

- **NULL dilution in the weighted average**: `SUM(metric * sample_count)` skips NULL products while `SUM(sample_count)` still counts those minutes. Docker and zfs fill all columns per row and proxmox NULLs are systematic per `entity_id`, which is a group key, so this should not bite in practice. Stated in the migration comment.
- Windows narrower than one bucket are skipped rather than refreshed: `refresh_continuous_aggregate` rejects them, and the refresh policy covers that tail anyway.
- The 1800s chart window cap is untouched, so the dashboard chart still cannot reach the aggregates.

## Testing

`bun run typecheck:all` clean. `src/lib/database/**`: 298 pass, 0 fail (includes the new `stats-rollup-backfill.test.ts`: real `min(time)` drives the windows, empty table skips to retention, retention lands only after both tiers, a failed window leaves retention unapplied without throwing and the next source still runs, a second process skips via the lock, re-runs stay `if_not_exists`).

**No TimescaleDB instance was available, so no migration and no backfill statement was executed.** The SQL remains unverified against a live server.

---
_Generated by [Claude Code](https://claude.ai/code/session_01E72xKQhMceYgcw4WrtjZau)_


src/lib/database/migrate.ts
```diff
@@ -16,6 +16,92 @@ const migrationsDir = join(projectRoot, 'migrations');
...
+  }
+
+  statements.push(current);
+  return statements.map((statement) => statement.trim()).filter((statement) => statement.length > 0);
+}
+
+function hasNoTransactionDirective(sql: string): boolean {
+  return sql.split('\n').some((line) => line.trim() === NO_TRANSACTION_DIRECTIVE);
+}
+
```
jaredglaser: this feels crazy. surely an abstraction for this already exists

jaredglaser: You are right, and I think the answer is better than swapping in a library: **this splitter should not need to exist.**

## Why it is here

It exists for exactly one reason. `CALL refresh_continuous_aggregate` cannot run inside a transaction block, and node-postgres sends a multi-statement string over the simple query protocol, which wraps the batch in an implicit transaction. So the statements have to be sent one at a time, which means the file has to be split. 030 has 22 such `CALL`s; every other statement in it is ordinary transactional DDL.

The splitter is a consequence of those `CALL`s being trapped in a `.sql` file. Move them out and the whole problem dissolves.

## The alternative

Do the backfill as a startup task instead of inside the migration, which is the same pattern just adopted for the git-token hash backfill in #279 (`startGitTokenHashBackfill` alongside `startDeployRecovery` in `server-init.ts`). TypeScript issues each `CALL` as its own `client.query`, so there is nothing to split, by construction.

That deletes, in order:

- roughly 70 lines of hand-rolled SQL lexer
- the `-- migrate:no-transaction` directive and its whole execution mode
- the "every statement must be individually idempotent because there is no rollback" hazard, which is the entire reason #369 needed a forced-replay check
- the known non-idempotent `ALTER MATERIALIZED VIEW ... SET (timescaledb.compress...)` problem at 030:156 and :161, since those move back inside a transactional file

The migrations go back to ordinary DDL. And it is arguably more correct: retention would be applied by the backfill task **after it confirms the rollup succeeded**, rather than relying on statement ordering within a file. Right now, ordering is the only thing standing between a failed backfill and a retention policy that deletes unrolled-up history.

It also buys real backfill behavior for free: windows computed from the actual `min(time)` rather than hardcoded literals, progress logging, and resumability across restarts.

## If you would rather keep it file-based

Then yes, libraries exist, and the right class is a **splitter, not a parser**: it only needs lexical structure, not grammar. `@databases/split-sql-query` is purpose-built for this, and `sql-query-identifier` (used by Beekeeper Studio and sqlectron) handles PostgreSQL dollar-quoting. I would avoid `node-sql-parser`, which actually parses and would likely reject TimescaleDB DDL like `CREATE MATERIALIZED VIEW ... WITH (timescaledb.continuous)`. `libpg-query` is the real PostgreSQL parser compiled to WASM and is correct by construction, but it is a multi-megabyte dependency for one job.

For what it is worth, the implementation here is better than most hand-rolled attempts: it handles `--` and `/* */` comments, single and double quotes with doubled-quote escaping, and dollar-quoted bodies with tags, which matters because these migrations do contain `DO $$ ... $$` blocks. Two gaps I did find:

- **Nested block comments.** PostgreSQL allows `/* /* */ */`; this takes the first `*/`, so the outer comment terminates early and the remainder leaks back into SQL.
- **`E''` escape strings.** Only doubled `''` is handled, so `E'foo\'bar; baz'` splits at the semicolon.

Neither is reachable from the current migrations, which is why it passes today. They are also exactly the kind of thing that makes hand-rolled SQL lexers a bad long-term bet.

My recommendation is to move the backfill to a startup task and delete the splitter along with the no-transaction mode. Happy to dispatch that if you agree.

---

jaredglaser: you should do the backfill as a startup task

jaredglaser: Done in `f6071b5`. The splitter is gone and `src/lib/database/migrate.ts` is byte-identical to its base again, confirmed by an empty `git diff` against `feat/f5-f6-timescale-tuning` for both it and its test file. `splitSqlStatements`, `NO_TRANSACTION_DIRECTIVE`, `DOLLAR_TAG`, `hasNoTransactionDirective` and `applyWithoutTransaction` are all removed; every migration runs in one transaction again.

The three migrations now carry zero `CALL refresh_continuous_aggregate`, zero `add_retention_policy` and no directive. They shrank from 174/151/184 lines to 135/119/152, keeping only the transactional DDL: chunk intervals, both `WITH NO DATA` aggregate definitions, the `_1h_avg` dividing view, refresh policies, and compression.

`runStatsRollupBackfill` in `src/lib/database/stats-rollup-backfill.ts` is wired into `initServer()` next to `startDeployRecovery` and `startGitTokenHashBackfill`. Per source it reads the real `min(time)`, refreshes `_1m` in bucket-aligned 2-day windows, then `_1h` in 6-day windows, and **only then** adds retention. A failure logs, leaves that source without retention and moves on, so the next boot retries. That inverts the previous risk: statement ordering inside a file was the only thing standing between a failed backfill and a retention policy deleting unrolled-up history.

Three details worth noting for review:

- **Windows are literal timestamps, not bind parameters**, so each `CALL` goes out as the sole statement of a simple-protocol query and stays outside a transaction block.
- **Advisory lock** is `pg_try_advisory_lock(0x686c6d32)` on a dedicated `PoolClient`, deliberately distinct from the migrations key so a long backfill never blocks a migrating process. A second holder logs and returns rather than blocking.
- **First-boot ordering** is handled: migrations run in the worker while this runs in the web process, so if the web reaches the task first, `to_regclass` finds no rollups, the source is skipped with an info log, and a later boot picks it up. No retention is in force meanwhile, so it fails toward keeping data.

The non-idempotent `ALTER MATERIALIZED VIEW ... SET (timescaledb.compress...)` stops mattering, since the file is transactional again and a mid-file failure rolls back.

One consequence for #369: its forced-replay check is now moot. It keys off a directive that no longer has an implementation, so it would discover zero files forever. Worth removing there rather than carrying dead code.

`typecheck:all` clean; `src/lib/database/` 298 pass, 0 fail. Still no TimescaleDB instance available, so none of this SQL has been executed against a live server.

---

Issue #416, opened 2026-08-02
Adopt Ansible as the execution layer for host and stack management


jaredglaser:
## Problem

Stack deployment today runs entirely through the agent sidecars. Compose definitions are
edited in-app, stored in the git-backed stack repo and `deploy_history`, and pushed to hosts
by `POST /stacks/deploy` on the agent. Two costs follow from that:

**The application is a second source of truth for deployed state.** Anything that changes a
host outside the app is invisible to it. There is no way to ask "does this host still match
what we think we deployed", and no way to reconcile when the answer is no. Issue #95 already
tracks drift detection as a gap, and #382/#383 are symptoms of trying to keep two records
agreeing.

**Only Compose is covered.** Docker engine installation, service users, directory ownership
and permissions, mounts, base packages, backup hooks, and LXC/VM lifecycle have no home in
the tool. Every one of those stays manual, which means the host a stack lands on is
configured by hand while the stack on top of it is managed by the app.

Stack management and agent functionality are already marked unstable and under active
development in the README, so changing the model now costs less than hardening the current
one and changing it later.

## Proposal

**Ansible becomes the write path. The agent remains the read path.**

| Concern | Owner |
| --- | --- |
| Metrics streaming, container/host state, logs | Agent (unchanged) |
| Compose stack deploys | Ansible |
| Host baseline (engine, users, dirs, mounts, packages) | Ansible |
| Proxmox guest lifecycle (future) | Ansible |
| Scheduling, secrets, UI, history | Homelab Manager |

To be explicit, because "rewrite to use Ansible" reads like a component deletion and this is
not one: **the agent is not removed.** It keeps `/stats/stream`, `/logs/:containerId`,
`/containers/events`, `/zfs/stats/stream`, `/zfs/pools`, `/health`, and `/info`. Everything
under `/stacks/*` and the container control endpoints move to Ansible over time. The worker
collector path, `docker_stats`, `docker_container_events`, `StatsPollService`, and every SSE
stats channel are untouched.

### Why Ansible rather than extending the agent

- One engine covers Compose stacks, host configuration, and later Proxmox guests. Extending
  the agent gives a Compose-shaped engine plus manual work everywhere else.
- `--check --diff` is preview-before-apply across every managed concern, for free. Building
  the equivalent by hand means reimplementing drift detection per resource type.
- `community.docker` and `community.proxmox` are maintained upstream. The alternative is
  hand-maintaining API surface coverage in the agent.

## Decisions

[...]

Issue #407, opened 2026-07-29
header: menu-route derivation fires authenticated queries on every route, hides failures, and races drawer expansion


jaredglaser:
Found while reviewing the mobile stack (#399-#403). All three symptoms come from one change (`ce82feb8`, "expand drawer submenus instead of navigating") and share a root cause, so they are filed together.

## Root cause

`useMenuRoutes()` in `src/components/header/menus.tsx` decides which routes own a submenu by counting query results:

```ts
export function useMenuRoutes(): ReadonlySet<MenuRouteKey> {
  const { data: stacks } = useStackSummaries()
  const { data: hosts } = useDockerHostNames()

  const routes = new Set<MenuRouteKey>()
  if (shouldRenderMenu(hosts?.length ?? 0)) routes.add(/docker)
  if (shouldRenderMenu(stacks?.length ?? 0)) routes.add(/stacks)
  if (shouldRenderMenu(SETTINGS_SECTIONS.length)) routes.add(/settings)
  return routes
}
```

`src/components/header/Header.tsx:36` calls it unconditionally, and `Header` renders in the root layout. Menu presence is now a function of async data that used to be a static flag (`item.hasMenu`).

### 1. Two authenticated server functions fire on every route load, and 401 before login

`src/components/AppShell.tsx:37` deliberately gates its preloads:

```tsx
if (authEnabled && !user) return;
```

with the stated reason that firing before the session check completes causes 401s. `useMenuRoutes` has no such gate. `useAuth` sets `loading: false` in its `finally` before `navigate({ to: /login })` lands, so `AppShell` renders `<Header user={null} />` for at least one commit, and `listStacks()` / `listManagedHostNames()` fire and 401, with react-query retries on top.

Previously these queries lived inside `StacksMenuContent` / `DockerHostsMenuContent` and only ran when a menu was actually opened, so an unauthenticated load never touched them.

### 2. A failed query silently degrades the nav with no signal

`grep -n "isLoading|isError|Failed to load|Loading" src/components/header/menus.tsx` returns nothing: the loading, error and empty branches were removed, and the six tests covering them were deleted from `menus.test.tsx`.

So a rejected `listStacks` leaves `data === undefined`, `useMenuRoutes` drops `/stacks`, and the tab silently becomes an ordinary link. Nothing anywhere tells the user the stack list failed to load. The nav looks deliberately configured rather than broken.

### 3. Drawer expansion races the queries

[...]

Issue #404, opened 2026-07-29
a11y: nav menus claim menu semantics without keyboard support


jaredglaser:
Found while reviewing the mobile stack (#399-#403). Deferred out of that stack because the cause predates it.

## Problem

The desktop nav popover and the mobile drawer submenus both wrap their contents in `role="menu"` with `role="menuitem"` children:

- `src/components/header/Header.tsx:156` (desktop popover container)
- `src/components/header/menus.tsx` (five sites: menu container plus menuitem links)
- `src/components/header/MobileNav.tsx` (drawer submenu container)

Nothing implements the keyboard contract those roles promise. There is no roving tabindex, no arrow-key handling, and the disclosure buttons carry no `aria-haspopup="menu"`.

The consequence is that assistive tech is actively misled rather than merely underserved. NVDA and JAWS switch to application/focus mode inside a `menu` and expect Up/Down to move between items with Tab exiting the whole menu; here arrows do nothing and Tab walks every item individually. VoiceOver announces "menu, N items" and then behaves like a plain list.

These are navigation links, not commands. The `menu` role is for a list of actions, so the roles are the wrong choice regardless of whether the keyboard handling gets written.

## Why this is not in the mobile stack

The roles arrive from #165 (`Header.tsx`) and #202 (`menus.tsx`), long before the mobile work. That stack adds no `role` attributes to `menus.tsx`; it only reuses the existing menu content inside the drawer. Fixing it therefore changes pre-existing desktop behavior, which does not belong in "make the app usable at phone widths".

It also cannot be fixed only in the drawer: removing the container `role="menu"` while `menus.tsx` keeps emitting `role="menuitem"` children leaves orphaned menuitems, which is worse than the current state. Both call sites have to change together.

## Proposed solution

Drop the menu roles rather than implement the keyboard pattern, since the content is navigation:

1. Remove `role="menu"` from `Header.tsx:156` and `MobileNav.tsx`.
2. Remove `role="menuitem"` from the links in `menus.tsx`.
3. Structure each menu as a `<ul>`/`<li>` list of links. The existing `aria-expanded` on the disclosure button plus a labelled list is the conventional disclosure-navigation pattern and needs no JS.
4. Keep `aria-haspopup` off the triggers, since they no longer open a menu.

## Blast radius

14 `menuitem` references across five files need requerying, three of them test files that assert `getByRole("menuitem", ...)`:

- `src/components/__tests__/header/MobileNav.test.tsx`
[...]

PR #417, opened 2026-08-02
feat(ansible): host-baseline execution layer behind a feature flag

jaredglaser:
Proof of concept for #416. Host baseline only: no Compose deploys, no Proxmox guests, no git-backed history, no rollback, no drift detection, no scheduling, and nothing removed from the agent.

## What and why

Ansible becomes the write path for host-level configuration while the agent stays the read path, unchanged. The agent keeps every endpoint it has, including `/stacks/*`, which is still the only way stacks are deployed. This adds a second, feature-flagged path that covers host baseline and proves the plumbing where a failed run leaves a host unchanged.

Off by default (`ANSIBLE_RUNNER_ENABLED`, absent means off). With the flag off there is no sidecar, the SSE route emits its error frame and closes, and the server functions throw before touching a network.

## Flow

```
/automation -> startAnsibleRun (admin-only) -> AnsibleRunService -> ansible-sidecar
     -> ansible-runner (run_async) -> managed hosts over SSH

ansible-sidecar SSE -> AnsibleRunService -> ansible_run_events
     -> ansibleRunBroadcastService -> /api/ansible-runs -> useAnsibleRun -> AutomationPanel

hlm_inventory.py --list -> GET /api/ansible-inventory (sidecar token) -> managed_hosts
```

The event path deliberately mirrors the one stack-status already takes: a producer, a persisted record, an in-process broadcast service, `createBroadcastSseHandler`, `useSseChannel`. The one difference is that there is no `LISTEN`/`NOTIFY` hop, because the web process that dispatches the run is the same one draining the sidecar stream.

## What else this touches

- **Migration 033.** Highest on `main` is 032 and no open PR adds one, so the number does not collide. Two new tables, no changes to existing ones.
- **`stack_secrets`** gains a reserved scope, `__ansible__`. No schema change (the table has no FK to stacks); variables in that scope become extravars of the same name. Nothing reads that scope except the Ansible path.
- **`src/lib/server-init.ts`** gains one startup task, which returns immediately when the flag is off.
- **`src/lib/mock/handlers/server-functions.ts`** gains three entries. `registry-coverage.test.ts` requires a mock for every `createServerFn` export under `src`, so the three new ones are registered; the demo mocks refuse to start a run, matching what a demo user would actually see.
- **`bunfig.toml`** ignores `AutomationPanelConnected.tsx` for coverage, same treatment as `ManagedHostsCardConnected.tsx`.
- `routeTree.gen.ts` regenerated for three new routes. Nothing outside the above reads the new tables, the new SSE channel, or the new env vars.

## Decisions I had to make that the brief did not specify

- **Where the tested logic lives.** Runner-event normalization and inventory generation are both TypeScript, not Python: the repo has no Python CI job, and putting them in the app keeps them under the existing coverage gate. The sidecar's Python is thin enough (request parsing, an event buffer, a per-host lock) that its absence from CI is not a coverage hole.
- **`suppress_env_files=True`, and its residual.** Without it `ansible-runner` writes extravars to `<private_data_dir>/env/extravars` in plaintext, which would violate the "plaintext never touches disk" constraint outright. With it, extravars are passed on the `ansible-playbook` argument vector instead, so a decrypted value is visible to anything that can read `/proc` inside the sidecar. That is the residual exposure and is why nothing else runs in that container. The alternative (pass secrets as `envvars` and read them with `lookup('env', ...)`) avoids argv entirely but is not extravars, so I kept the specified interface and documented the tradeoff. Worth a decision before this goes past POC.
- **`ANSIBLE_HOST_KEY_CHECKING`.** `ansible-runner` sets it to `False` when the variable is absent from the environment it is handed, so verification is off unless you turn it on. The sidecar passes it explicitly, defaulting to `True`, with a `known_hosts` file on a volume. Seeding that file at enrollment is not built here.
- **Self-management, resolved as exclude.** The host named by `ANSIBLE_SELF_HOST_NAME` gets `hlm_self_managed: true` in its hostvars and the baseline play ends early on it. The brief asked for an explicit choice; accepting that `docker compose up -d` may kill the process that issued it trades a reliable outcome for a convenience, and the existing stuck-deploy recovery exists to clean up crashes rather than to normalize them.
- **In-process concurrency guard, not an advisory lock.** There is one sidecar, so a process-local host set is sufficient and honest. A second sidecar would need the PostgreSQL advisory lock the deploy pipeline already uses for stacks.
- **`/automation` is not linked from navigation.** Adding a nav entry would need the client to know a server-side flag, and the nav derivation is actively being reworked (#404, #407). The route is reachable by URL while the feature is behind a flag.
- **Allowlist over denylist for event fields.** The sidecar strips `event_data.res` and `task_args`, but the decision about what reaches the database and the browser is an allowlist in `normalizeRunnerEvent`, so a future runner field cannot leak by default. There is a test asserting a secret planted in `res` does not survive normalization.
- **Sidecar has no durable state.** Runs are recorded by the dispatching web process, and startup fails stranded `pending`/`running` rows, because a dead dispatcher leaves a run with no writer.

## Discrepancy with the brief

The brief said the README lists OIDC auth as landed-but-undocumented behind `AUTH_ENABLED`, and that this feature must be gated behind completed authentication. That is not the current state: auth is required by default, `AUTH_DISABLED=true` is the explicit opt-out, `enforceAuthConfig()` fails startup on the removed legacy `AUTH_ENABLED`, and setup is documented in the self-hosting guide. So there was nothing to wait on. The gating is still applied as intended: `requireRole('admin')` on all three server functions, `authenticateSSE` on the stream, and a separate machine token (not a session cookie) on the inventory endpoint, which is read-only and cannot trigger a run.

## Acceptance criteria

| Criterion | Where |
| --- | --- |
| Multi-host run, one host offline, per-host outcomes independent | `any_errors_fatal: false` + `ignore_unreachable: true` in the playbook; `summarizeStats` reads `playbook_on_stats.dark` as its own `unreachable` outcome; `statusFromSummaries` does not fail a run for it. Covered by `run-service.test.ts` ("reports an unreachable-only run as succeeded") and `events.test.ts`. |
| `--check --diff` exposed, previews without mutating | `checkMode` on the server function, through `SidecarClient.startRun` to `cmdline="--check --diff"`. Surfaced in the UI as "Preview only", defaulted on in the connected component. |
| Task-level events stream to the UI in near real time | `event_handler` on `run_async` -> sidecar SSE -> `AnsibleRunService.consume` -> broadcast -> `/api/ansible-runs`. `useAnsibleRun` renders play, task, and per-host lines as they arrive. |
| No secret in logs, artifacts on disk, or the event stream | `suppress_env_files=True` (nothing on disk), `no_log: true` on the secret-consuming task (nothing in stdout or `event_data.res`), allowlist normalization (nothing on the wire). Residual argv exposure noted above. |
| Tests for event parsing and inventory generation | `src/lib/ansible/__tests__/events.test.ts` and `inventory.test.ts`, plus the sidecar client, run service, broadcast service, inventory handler, repository, handlers, schemas, channel, hook, component, and both API routes. |

## Verification

Run against `.bun-version` (1.3.14), not the sandbox default:

- `bun run typecheck:all` clean.
- `bun run test` (homelab-manager, with the coverage gate): **3614 pass, 0 fail**. Lines 98.49% against a 98% floor, functions 95.26% against 94%. Both moved up.
- `bun run typecheck:agent` clean; agent tests 351 pass, 0 fail.
- `bun run build` succeeds.
- Python: `py_compile` on all three modules; direct checks of `parse_start_request` (accepts a valid body, rejects a path-shaped playbook, a traversal `runId`, an empty host list, and a non-object `extravars`), of the `event_data` strip, and of the run event buffer replaying to a late subscriber. `hlm_inventory.py` exercised for `--list`, `--host`, and no-args.

Two things I did **not** verify, stated plainly:

- **E2e did not run here.** The sandbox's Playwright browser directory does not contain the headless-shell revision this repo's pinned Playwright expects, and the environment forbids re-installing browsers. CI installs its own, so this needs to be read off the pipeline. No e2e spec covers `/automation` and no existing spec should be affected, but that is reasoning, not a green run.
- **No real host was involved.** No SSH, no sidecar container, no `ansible-runner` process. Every runner-event assertion is against fixtures shaped from the documented event schema. The multi-host and offline-host behaviour is verified at the normalization and run-service layer, not end to end against two machines.

Coverage differs between local and CI, so cite the pipeline for the numbers above rather than this description.

Closes nothing. Increment 1 of the plan in #416.

---
_Generated by [Claude Code](https://claude.ai/code/session_01S5Yg4QVYde8kyDDsDiZJEe)_

## Summary by CodeRabbit

* **New Features**
  * Added an optional Automation page for administrators to select managed hosts and run host-baseline checks.
  * Added live task output, run status, cancellation, check mode, and per-host outcome summaries.
  * Added authenticated inventory access and real-time run updates.
  * Ansible execution is disabled by default and configurable through environment settings.

* **Documentation**
  * Documented setup, configuration, security considerations, supported behavior, and architecture.

* **Bug Fixes**
  * Added recovery for interrupted runs after server restarts.


jaredglaser:
@coderabbitai review

